PR #74, opened 2017-11-14
Change .gitignore to add composer.lock in the project. Change Localis…

marcospaegle:
Change .gitignore to add composer.lock in the project. Change Localisation to Localization on en-GB/settings.php

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,3 @@ npm-debug.log
 robots.txt
 _ide_helper.php
 .phpstorm.meta.php
-composer.lock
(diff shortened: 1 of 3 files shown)


denisdulici:
Thanks Marco. Any reason why should add composer.lock to project? In fact we don't add even composer.json to the end-user package https://github.com/akaunting/akaunting/releases

As for Localisation, the default language of Akaunting is en-GB, not en-US so such words are used to be with s, not z. If you wish here you can contribute to en-US translation https://crowdin.com/project/akaunting

PR #91, opened 2017-11-22
Add `db_charset` to fix postgresql installation error.

iquad:
Need to set `db_charset` and use it for installation connection to contest `missing index: 'charset'` error. Also need to add `DB_CHARSET` to `.env` file. 

Error happens on postgresql 10 (and probably previous versions).


denisdulici:
Thanks Mert

PR #120, opened 2017-12-04
Added Docker containerisation + CLI installation

carvallegro:
- Dockerfile and docker-compose file to create a container out of the Akaunting app. 
    Some docker-compose example files can be found in the `docker` folder.
- A CLI Artisan command allows to do the installation process in one go without having to go through the UI process. For more informations, do : `php artisan help app:configure`


carvallegro:
Yeah I've thought about it. I think the first step is to create the akaunting image and then use a `docker-compose` to link the two. I didn't want to make the image linked to any specific DB. I'd like to find a way to set up automatically the DB in Akaunting. I'll investigate during/after Christmas time.

Still needs to be tested depending on the environment but it's a good way to have a local env up and running.

I've also talked with @denisdulici about pushing the image to DockerHub, but an official Akaunting account needs to be created first.

carvallegro:
@beinbm Could you please review the most recent changes?

docker/docker-compose.build.yaml
```diff
@@ -0,0 +1,7 @@
+version: '3'
+services:
+  web:
+    image: akaunting
+    build: ../
```
beinbm: you dont need a second build yaml file. just add this build line additionally into the other yml file

Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM php:apache
+
+COPY . /var/www/html/
```
beinbm: can you move the _copy_ and _chmod of the project files_ below apt-get?
like here: https://github.com/akaunting/akaunting/pull/162/files#diff-3254677a7917c6c01f55212f86c57fbf

docker/docker-compose.build.yaml
carvallegro: @beinbm I'm keeping this file because Dockerfiles don't support image tagging. This way it'll be easier to tag the version in the future

Dockerfile
carvallegro: Done

docker/docker-compose.build.yaml
beinbm: https://docs.docker.com/engine/reference/commandline/build/#tag-an-image--t

denisdulici:
There is no need for `composer dump-autoload` command anymore https://github.com/akaunting/akaunting/commit/3c49b8ce1bed1cb46b666c030f6baafa3a1f0839

Are we ready to go?

carvallegro:
@denisdulici  We're ready to go. I think it's a good first version that we can merge.
I'll discuss with @beinbm about future improvements. I'm thinking

- DB migration 
- Embed the install CLI into the the container startup
- ...

denisdulici:
Thanks @carvallegro @beinbm

PR #162, opened 2017-12-22
improved Dockerfile & added database service in docker-compose.yml

beinbm:
I improved the docker container of @carvallegro and added mysql as a database service


emrahcetiner:
@beinbm  "composer install" command can be run in dockerfile? in this way all dependencies are installed in docker image.

carvallegro:
I wouldn't add the database in the Dockerfile. The advantage to stay database agnostic is to be able to add any kind of database Akaunting is compatible with.

I'd rather use the `mysql:latest` image in the docker-compose only.

beinbm:
@emrahcetiner, i did that in my first attempt, but later saw a recommendation not to do it. 
https://github.com/docker-library/php/issues/344

@carvallegro, it is like that in my commit. In the Dockerfile I just load pdo_mysql php extention

carvallegro:
Ah right! I didn't know what it was. Maybe it's worth adding PHP extensions for other databases?

carvallegro:
@beinbm , I've updated my PR with some major changes and I've taken your changes in consideration

denisdulici:
Is this ready to merge Valentin?

beinbm:
how is it possible that I add commits to your MR? I pulled your master, changed based on my comments and commited here again. So this MR is a mix of both of our code.
Maybe we should work in the same repository :)

carvallegro:
Hello @beinbm  I think you need to create a PR to my repo : https://github.com/carvallegro/akaunting 
I'll add you as a Collaborator right away :)

And yes, we definitely should to avoid confusion

carvallegro:
@beinbm I've added you as a Collaborator on the project. You should get an invite. I've set the default branch protection on the repo (PR creation and reviewal)

PR #186, opened 2018-01-15
Euro - Symbol Position

DavidLiedke:
Before --> After


denisdulici:
Thanks David but `Before` is the default position:
http://publications.europa.eu/code/en/en-370303.htm#position

You may edit as you wish from `Settings > Currencies` page.

PR #215, opened 2018-02-12
Patch2 add validations

mbdwey:


denisdulici:
Can you plz provide instructions on how to reproduce these errors?

mbdwey:
Not sure but check migrate:refresh and missing ENV variables  scenarios

denisdulici:
That command is not applicable for Akaunting as it has a separate installation process.

PR #217, opened 2018-02-13
Patch3

mbdwey:
Prevent duplicate module install 
Pass console reference throw the even


denisdulici:
Mohamed, what happens in duplicate install?

PR #313, opened 2018-04-27
Added lang attribute to layout views using the default localisation

kostasdizas:
The lang attribute helps browsers render special language features better


denisdulici:
Thanks Kostas

PR #320, opened 2018-04-30
Invoice/Bill: Decimal value for quantity

denisdulici:


EgyptianM:
Waiting for decimal value for price as it's essential for accounting.

PR #374, opened 2018-06-07
Build with a separate command

rodolfovillaruz:


birkof:
What's the difference between that and `docker-compose up -d --build`?

rodolfovillaruz:
I believe it's the same. We could use that.

denisdulici:
Guys, is this ready to merge?

denisdulici:
Thanks guys.

Issue #275, opened 2018-03-19
Jessenger/date Error on composer update


kwhohasamullet:
Please, submit only real issues. Use the forum for support, feature requests, proposals, new versions, help etc. https://akaunting.com/forum

### Steps to reproduce the issue
1. Run composer update on akaunting installation

### Expected result
Composer update the complete as normal

### Actual result
Console error
----------------------------
In Date.php line 10:

  Declaration of Jenssegers\Date\Date::diffForHumans(?Carbon\Carbon $since =
  NULL, $absolute = false, $short = false) should be compatible with Carbon\C
  arbon::diffForHumans(?Carbon\Carbon $other = NULL, $absolute = false, $shor
  t = false, $parts = 1)

### Additional comments
I was able to resolve this updating jessenger/date in composer.json to 3.3.*
I am not currently able to submit a pull request but hopefully it is easy enough for someone else to do
Further info on this error can be found here https://github.com/jenssegers/date/issues/280

denisdulici:
Thanks Kieran.

PR #447, opened 2018-08-07
Update popover function call method signature

shreyansh9226:
Currently on line #275
`$('a[rel=popover]').popover({ html: 'true',`

html true where `'true'` is within quotes, Which causes issue when trying to open popover. As popover expects Boolean. As mentioned at: https://getbootstrap.com/docs/4.0/components/popovers/#options

### DIFF
--- a/resources/views/expenses/bills/create.blade.php
+++ b/resources/views/expenses/bills/create.blade.php
@@ -273,7 +273,7 @@
             });
 
             $('a[rel=popover]').popover({
-                html: 'true',
+                html: true,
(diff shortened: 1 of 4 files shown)


cuneytsenturk:
Thanks @shreyansh9226 

Could you fix those pages Bill, Invoice create and edit template,

I'll merge if you send those cahnges.

shreyansh9226:
@cuneytsenturk

cuneytsenturk:
Thank you @shreyansh9226

PR #456, opened 2018-08-23
README fix for docker-compose.yml file

deleonjavier:
When looking into installing akaunting via docker I noticed that the location for the `docker-compose` file was incorrectly linked. 

Small fix but a fix 🎉


denisdulici:
Thanks Javier :wink:

Issue #171, opened 2017-12-28
After download invoice, the over line text not correct


asyrafmuda:
I create an invoice, after that, i want to download. I see the over line text in invoice are not correct. (as attachment). 
![screen shot 2017-12-28 at 12 11 49 pm](https://user-images.githubusercontent.com/11627478/34400125-3252b0b8-ebc9-11e7-9d1c-36c0f046f6a1.png)

After reviews, when i try to print out, just nice. right now, i just need to click print, and save as pdf. it working. 

thanks. btw, your website discussion, i think they is wrong when i want to open the discussion.

denisdulici:
It seems that DomPDF doesn't read the output correctly 😞

angelscabrerag:
@denisdulici can you guide me which files need to edit to try to solve this? I see akaunting is using the barryvdh DomPdf libary, so i need some points to start.

denisdulici:
The problem is with [DomPDF ](https://github.com/dompdf/dompdf) library as it seems to handle just a few CSS3 properties. There is no exact file to edit.

angelscabrerag:
Im was playing around, made some css fixes but still overflowing the divs with class "col-xs-5" (company info, invoice info & payment info).

lalogarcia123:
For that problem, I print as a PDF in Chrome, and unselect the header and footer

PR #540, opened 2018-10-10
Fix invoice print layout and PDF

developerbmw:
I've changed the invoice view so that it is a standalone view, not including all the normal CSS / JS etc as that slows down the PDF massively. Check out how fast the PDF download is now...

Also I've completely re-done the layout using a combination of inline-block and tables which are supported by DOMPDF so now the PDF matches the print web page.

### DIFF
--- a/resources/views/incomes/invoices/invoice.blade.php
+++ b/resources/views/incomes/invoices/invoice.blade.php
@@ -2,139 +2,217 @@
 
 @section('title', trans_choice('general.invoices', 1) . ': ' . $invoice->invoice_number)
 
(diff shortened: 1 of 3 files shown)


developerbmw:
Fixes #171

denisdulici:
Brett, we must keep the current structure in order to make the system extendible so instead of deleting, please apply the changes to layout + partial files and include them.

developerbmw:
Thanks Denis, I have updated the commit

denisdulici:
Perfect Brett :wink:

PR #585, opened 2018-11-02
Fixed bug in reconciliations that prevented it from working in linux and other OSes

### LINKED ISSUE #578: Reconciliations don't work on non-windows environments
The new reconciliations module uses the name of the expected model classes to, at the controller level, decide where to find the contact name, and at the reconciliation views, to show the amount as a withdrawal or deposit. The code seems to be using the `basename` method to get the class name out of the qualified class string. `basename` can only be used in this way on windows environments as it is meant for paths and windows contain backslashes (same as the qualified class names). It should instead use `is_a`. Additionally, it would be better if the transaction could decode the class name to something the views can use to draw accordingly, without having to know about the underlying class name.

### Steps to reproduce the issue

1. Install akaunting on linux
2. Create one revenue and one payment transaction
3. Create a reconciliation for the account used above

### Expected result

The amount from the revenue transaction shows as a deposit and the amount from the payment shows as a withdrawal.

### Actual result

Both transactions show as withdrawals

### System information (Akaunting, PHP versions)

Akaunting 1.3-dev, PHP 7.0

kostasdizas:
Fixes #578

Maybe not the most elegant fix, but it works. Ideally the four classes should share some common interface that allows better control and cleaner logic in lookups, saves, and calculations.

### DIFF
--- a/app/Http/Controllers/Banking/Reconciliations.php
+++ b/app/Http/Controllers/Banking/Reconciliations.php
@@ -218,7 +218,7 @@ class Reconciliations extends Controller
             $m::where('account_id', $account->id)->whereBetween('paid_at', [$started[0], $ended[0]])->each(function($item) use(&$transactions, $model) {
                 $item->model = $model;
 
-                if ((basename($model) == 'Invoice') || (basename($model) == 'Revenue')) {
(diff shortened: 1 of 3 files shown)


cuneytsenturk:
Thanks @kostasdizas

PR #632, opened 2018-11-17
Added composer.phar

skay1994:
Added file to run composer commands on servers with no local installation to run composer.


denisdulici:
Thanks Jorge but this is not needed.

PR #634, opened 2018-11-17
Remove check updates from modules in dev

skay1994:
Added check for modules not found in the sever, for modules in local development or modules created by user.


denisdulici:
Thanks Jorge but this is not needed.

Issue #631, opened 2018-11-17
Error 500 on adding new item to invoice after Update 1.3.2


fellek:
Its not possible to add items to invoice or bill.
Error appears after Update to Akaunting 1.3.2.

### Steps to reproduce the issue
1. Create new Invoice or Expenses Bill
2. Fill first Item Row
3. Add new Item

### Expected result
New Item Row should appear

### Actual result
No Item Row appears. 
Request: /incomes/invoices/addItem?item_row=1
Response: Statuscode 500 Internal Server Error

### System information (Akaunting, PHP versions)
Akaunting 1.3.2
PHP 7.2.11
Ubuntu 16.04
Nginx

### Additional comments
Thanks for your codes!

JvanEst:
I have this exact same problem. Found the problem as an variable is tried to be initialised at a stage that the dropdown for the currency code is not yet present. I proposed a fix in my pull-request.

cuneytsenturk:
Solved here #635

cuneytsenturk:
Thank you feedback @fellek

PR #635, opened 2018-11-19
#631 Fixed: Add row button on Invoice and Bill

JvanEst:
Move the `currency_code` variable in a `$(document).ready` block. 

Currently the `currency_code` variable will always be empty. This is because the script is inserted in the head of the page and the `currency_code` dropdown is not available at that stage. By moving the `currency_code` variable to the document ready section the variable will be filled when the page is loaded.


cuneytsenturk:
Thanks @JvanEst

Issue #637, opened 2018-11-20
Discount Popover not working


JvanEst:
Please, submit only real issues. Use the forum for support, feature requests, proposals, new versions, help etc. https://akaunting.com/forum

### Steps to reproduce the issue
Create or edit an invoice, try to add a discount

### Expected result
Discount popover

### Actual result
Nothing.

### Additional comments
Again same issue as #631 the popover is initialised before the rest of the dom is ready.

I've proposed another pull request to fix this issue by moving the popover initialisation to the document ready. A neater fix would be to move all custom script to the bottom of the page. But I think that would be to much of a refactor at this time.

cuneytsenturk:
fixed here => https://github.com/akaunting/akaunting/pull/638

PR #638, opened 2018-11-20
#637 Fixed: Moved popover initialisation to Document ready

JvanEst:
Move the popover initialisation inside the `$(document).ready` block. This is will make sure that the element that the popover is initialised on is actually available in the DOM.


cuneytsenturk:
thanks @JvanEst

PR #654, opened 2018-11-29
No need to mount a folder you already copied in the image

h4wkmoon:
Plus, it generated access errrors.


beinbm:
the idea of the docker-compose file is to have a local dev environment. therefor a mount of the hosts folder where you work in is needed. 
I agree that we would not need to copy the files in the first place

denisdulici:
Docker moved to its own repo.

PR #671, opened 2018-12-09
Unique invoice link for guests

burakcakirel:
refs #554 Does not work for guests. Company ID should be added to querystring


cuneytsenturk:
Thanks for report @burakcakirel .

PR #681, opened 2018-12-14
Support for financial year customisation

sausin:
Instead of the current default financial year start from 01 Jan, this allows for customized start of the financial year, for eg. from 01 April (in case of India).


app/Http/Controllers/Common/Dashboard.php
```diff
@@ -33,6 +37,7 @@ class Dashboard extends Controller
     public function index()
     {
         $this->today = Date::today();
+        $this->financial_start = $financial_start = Date::parse(Setting::where('key', 'general.financial_start')->first()->value)->format('Y-m-d');
```
denisdulici: Find: `Setting::where('key', 'general.financial_start')->first()->value`
Replace: `setting('general.financial_start')`

app/Http/Controllers/Common/Dashboard.php
```diff
@@ -55,7 +60,8 @@ public function index()
             'donut_expenses',
             'accounts',
             'latest_incomes',
-            'latest_expenses'
+            'latest_expenses',
+            'financial_start'
```
denisdulici: This is also not needed as you can use the `setting` helper in views too.

sausin: Thanks for your suggestions! I'll update the code to make it neater.

I didn't look at the reports, so checking those now.

app/Http/Controllers/Common/Dashboard.php
sausin: Will update to use the helper

sausin:
@denisdulici please check now

database/seeds/Settings.php
```diff
@@ -28,6 +28,7 @@ private function create()
 
         Setting::set([
             'general.date_format'               => 'd M Y',
+            'general.financial_start'           => '01 Jan',
```
denisdulici: What happens for other languages?

sausin:
Using Date function now so it should be fixed. Let me know if anything else

denisdulici:
I get the following error when I change the fiscal start (i.e. 01 April) and try to access any report.

screenshot_1545487935

sausin:
@denisdulici Thanks for checking and sorry I missed a change in the report files. It should be good now.

I assume `Tax Summary` and `Profit & Loss` are working properly for you without the missing modifications.

denisdulici:
Everything seems to be working fine now, thanks for your contribution :clap:

sausin:
@denisdulici glad to be of help :+1:

PR #689, opened 2018-12-20
Fix calculation to show correct amount in modal

sausin:
Fix `grand_total` calculation. No $paid variable existed


sausin:
While I've updated the code to not change the method, I think the `if` condition is not really useful. The method in the class will return `0` if there is no paid amount, making the `if` condition redundant

cuneytsenturk:
Thank you @sausin

sausin:
:+1:

PR #699, opened 2018-12-25
Remove passing of unused variable

drbyte:
In PHP 7.3, calling `compact` on variables that aren't declared will throw an exception:

```
[2018-12-25 17:17:23] production.ERROR: ErrorException: compact(): Undefined variable: emails in /app/Http/Controllers/Incomes/Customers.php:32
```

### DIFF
--- a/app/Http/Controllers/Incomes/Customers.php
+++ b/app/Http/Controllers/Incomes/Customers.php
@@ -29,7 +29,7 @@ class Customers extends Controller
     {
         $customers = Customer::collect();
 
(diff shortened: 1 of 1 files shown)


drbyte:
The build failed because of an apostrophe in the customer name being checked in the test. Re-running the 5.6 build will show it to be fine.

(One could also update the tests to look for the escaped version of the customer name being tested for.)

cuneytsenturk:
Thanks @drbyte

PR #714, opened 2019-01-02
The switch button "Show to customer" is always On

enimiste:
- Akaunting version : 1.3.8
## Steps to reproduce
- Settings > Offline Payments
- Edit an existing row and make it not be shown to the customer
- Save
- Edit again, it is supposed to be No. But it is alawys Yes in the IHM.

## DB :
In the data base the data is correct : "customer":0 the same in the response of the ajax call

## Comments
In the interface of "Offline Payments" the radio button is always On even if the "customer" attribute is 1.


denisdulici:
Thanks Enimiste.

Issue #718, opened 2019-01-06
Dashboard total profits are displaying wrong value


bluecyclone:
Please, submit only real issues. Use the forum for support, feature requests, proposals, new versions, help etc. https://akaunting.com/forum

### Steps to reproduce the issue

Setup a fresh install
Create two bank accounts, one of them in USD and another one in EUR currency
Add $1 revenue to the first Bank Account
Transfer $1 USD to EUR Bank Account ( assume it has about 0.88 rates so 1USD should be 0.88 EUR )
Then add €0.88 EUR expenses from the second account. Now we have nothing in both of accounts
Now you see Total profit is calculating to 0.12 which should be 0

### Expected result

Total profits should be 0

### Actual result

Total profit is 0.12

### System information (Akaunting, PHP versions)

Akaunting 1.3.8
PHP 7.2
MySQL (MariaDB)

### Additional comments

It seems it is calculating only in USD and didn't check other currencies so It calculated ( 1 - 0.88 = 0.12 ) which should be ( $( 1*1 ) - $( 0.88 * 1.14 ) = $0

luizvid:
Hello, how are you?
I sent a pull request which I believe corrects this issue.
Thank you!

PR #729, opened 2019-01-15
Fixed issue #718

luizvid:
Fixed displaying wrong values when different currencies are used which resolves issue #718.
Also made some changes to conform with psr-2


cuneytsenturk:
thanks @luizvid

PR #741, opened 2019-01-28
Improved presentation for smaller screens

sausin:
The payments and histories being stacked side-by-side works well for medium screen upwards, but on small devices (phones) the view makes it rather difficult to read.

This change improves legibility on mobile gadgets.


cuneytsenturk:
Thank you @sausin.

PR #751, opened 2019-02-03
Closes #688

### LINKED ISSUE #688: Issues with inclusive tax calculation
### Steps to reproduce the issue
1. Create two tax rates (inclusive type, for eg 5%)
2. Create a bill (for eg. $110)
3. Add two inclusive rates
4. See tax calculation

### Expected result
Expect to see tax of $10 in total ($5 corresponding to each tax slab) and subtotal of $100

### Actual result
Only half the tax is calculated and that too is incorrectly calculated.

### System information (Akaunting, PHP versions)
Akaunting - 1.3.4
PHP - 7.2.10

### Additional comments
image

The tax calculation should sum up all the rates for inclusive type and calculate the base price. The taxes should then be calculated over this base price and also be summed together.

sausin:
The additional point about restricting the types of taxes is not covered here but the inclusive tax calculation is fixed to work properly

### DIFF
--- a/app/Http/Controllers/Common/Items.php
+++ b/app/Http/Controllers/Common/Items.php
@@ -358,37 +358,12 @@ class Items extends Controller
                     }
 
                     if ($inclusives) {
-                        if ($discount) {
-                            $item_tax_total = 0;
+                        $item_sub_and_tax_total = $item_discount_total + $item_tax_total;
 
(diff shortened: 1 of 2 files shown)


sausin:
@denisdulici if you could please review

cuneytsenturk:
Thank @sausin.

PR #757, opened 2019-02-05
Limiting selectable taxes based on type

sausin:
Issue was described in #688 that based on the way taxes are calculated, inclusive and other types are not compatible for simultaneous use.

This PR tries to address this by UI adjustments.


sausin:
It would be ideal if the other taxes could be hidden, but I couldn't find a way to make it work with select2

sausin:
@cuneytsenturk thoughts on this please?

denisdulici:
Please, resubmit this according to 2.0 version.

Issue #251, opened 2018-03-06
Configurable invoice, email and subject


ChipwizBen:
As per: https://akaunting.com/forum/discussion/general/invoice-email-to-customer

The subject is just 'Invoice' which is a great way to get the email caught in a spam filter. Is this / can this be configurable? Possibly even 'Invoice from <Company>'?

Secondly, the line of text in the email doesn't make sense:
"You are receiving this email because you have an upcoming $57.50 invoice to <customer name> customer."

Why would a customer receive an email that says they owe money to themselves?
 

Thirdly, the PDF is out of shape. All the tables are messed up and out of line or overlapping each other. The invoice is also 2 pages long while the text and boxes only span for half of the first page. Interestingly, the format when printing the PDF is perfectly OK.
 

And finally, when you click the "Pay Now" button in the email, it takes you to the login page for Akaunting. This customer doesn't have an account on the system that they're able to login to, and there doesn't appear to be a way to:    

- Change the email format to make it clearer
- Add payment instructions to the invoice

denisdulici:
As said in forum, the invoice email is going to be replaced by the Email Templates feature in the next version.

Marcel40625:
any updates?

quangbahoa:
+1

PR #771, opened 2019-02-21
changed invoice notification message text

kerrsmith:
changed message text so it is clear the email contains an attached invoice

### DIFF
--- a/resources/lang/en-GB/invoices.php
+++ b/resources/lang/en-GB/invoices.php
@@ -65,7 +65,7 @@ return [
     ],
 
     'notification' => [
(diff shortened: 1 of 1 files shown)


kerrsmith:
I saw in https://github.com/akaunting/akaunting/issues/251#issuecomment-371417905 that you are working on a new email template feature - this change is really just a basic fix to the email text wording so it is clearer what the email actually is until your new code has been implemented.

denisdulici:
Thanks for the PR but the new message is not exactly how it should be and as you said, a new email template feature is coming so closing this one.

PR #776, opened 2019-02-25
fixed validation bug

alanbrande:
Several validation checks where being done against $this->dbPort instead of the matching attribute.


cuneytsenturk:
thanks @alanbrande .

PR #790, opened 2019-03-15
Update index.blade.php

Douglasproglima:
Update for Account Balance.
setting('general.default_currency');


Douglasproglima:
Need to update this file, currency symbol is incorrect.

denisdulici:
Each account has it's own currency code so there is no problem there.

Issue #788, opened 2019-03-12
Invalid Pagination Path


kbhagchandani:
Pagination path is not being generated properly in sub components like customers.

### Steps to reproduce the issue
1. Go to Incomes -> Customers
2. Click on customer name. (You must be having 26 transaction for the selected customer. In my example I am marking customer id **1**)
3. Click on the page 2 link on the transactions list.

### Expected result
The link must be: _"incomes/customers/**1**/?page=2"_

### Actual result
Link being generated: _"/?page=2"_

### System information (Akaunting, PHP versions)
Akaunting Version 1.3.14
PHP Version 7.3

denisdulici:
Please, try this again with 2.0 version.

PR #795, opened 2019-03-23
#788 Bug Fix

kbhagchandani:
Fixes the pagination path


kbhagchandani:
composer.json is updated to utilize new packages which requires minimum of php version 7.1 to function properly which makes build failure. changes of composer.json can be discarded if previous unsuppported versions are planned to be supported further. Please guide.
http://php.net/supported-versions.php

denisdulici:
Please, remove the changes not related to pagination issue.

PR #797, opened 2019-03-23
PostgreSQL compatibility

shahalpk:
Fixed the issue with postgresql compatibilty in DB migrations & Support for Postgres in Installer


database/migrations/2018_06_30_000000_modify_enabled_column.php
```diff
@@ -12,6 +12,14 @@ class ModifyEnabledColumn extends Migration
      */
     public function up()
     {
+        
+        if ( DB::connection()->getPDO()->getAttribute(PDO::ATTR_DRIVER_NAME) == 'pgsql' ){
```
denisdulici: Why not use `env('DB_CONNECTION')`?

resources/lang/en-GB/install.php
```diff
@@ -28,6 +28,9 @@
         'username'          => 'Username',
         'password'          => 'Password',
         'name'              => 'Database',
+        'driver'            => 'Driver',
+        'port'              => 'Port',
+        'prefix'            => 'Table Prefix',
```
denisdulici: Please, remove prefix input as it's not related to this PR.

resources/views/install/database/create.blade.php
```diff
@@ -3,13 +3,20 @@
 @section('header', trans('install.steps.database'))
 
 @section('content')
+
+    {{ Form::selectGroup('driver', trans('install.database.driver'), 'plug', ['mysql' => 'MySQL', 'pgsql' => 'PostgreSQL'], old('driver', 'mysql'), ['required' => 'required'],  'col-md-12') }}
```
denisdulici: Must add other databases as option https://laravel.com/docs/5.4/database#introduction

Also, put it after hostname input.

app/Utilities/Installer.php
```diff
@@ -147,19 +148,28 @@ public static function createDbTables($host, $port, $database, $username, $passw
      * @param $database
      * @param $username
      * @param $password
+     * @param $driver
      *
      * @return bool
      */
-    public static function isDbValid($host, $port, $database, $username, $password)
+    public static function isDbValid($host, $port, $database, $username, $password, $driver)
```
hmert: `$prefix` should be added here too.

app/Utilities/Installer.php
```diff
@@ -147,19 +148,28 @@ public static function createDbTables($host, $port, $database, $username, $passw
...
         Config::set('database.connections.install_test', [
             'host'      => $host,
             'port'      => $port,
             'database'  => $database,
             'username'  => $username,
             'password'  => $password,
-            'driver'    => env('DB_CONNECTION', 'mysql'),
-            'charset'   => env('DB_CHARSET', 'utf8mb4'),
+            'driver'    => $driver,
+            'charset'   => $charset,
```
hmert: `$prefix` should be added here too.

resources/lang/en-GB/install.php
hmert: @denisdulici I think it is related. If someone wants to enable all databases, then it should be a separated PR. 
 
And that PR should consider support the _connection string_ usage. https://www.doctrine-project.org/projects/doctrine-dbal/en/2.9/reference/configuration.html

resources/views/install/database/create.blade.php
hmert: @denisdulici this PR focuses only PostgreSQL.

app/Utilities/Installer.php
shahalpk: @hmert isDbValid is only checking whether the connection parameters are proper. Hence `$prefix` can be ignored here.

app/Utilities/Installer.php
shahalpk: @hmert for testing db connectivity, i guess it doesn't make sense to pass prefix here. Let me know if there is a use case where the connectivity will fail for any different prefix.

database/migrations/2018_06_30_000000_modify_enabled_column.php
shahalpk: I have a feeling that this will be more accurate as its coming directly from the PDO driver. I think DB_CONNECTION is a user config and its better to rely on standard classes to get the actual connection type.

shahalpk:
Hi @denisdulici and @hmert, thanks for your review. please go through my comments and let me know if any changes are required

denisdulici:
My comments are still valid. Also, changes done to installer utility breaks the installation from console. Check out the `install` command.

app/Utilities/Installer.php
hmert: I was wrong about `$prefix` _libpq_ does not need a prefix to check connection is valid or not. https://www.postgresql.org/docs/10/libpq-connect.html

app/Utilities/Installer.php
hmert: And, surprisingly, there is any connectivity limitation about prefix.

hmert:
@shahalpk I've added my comments.

denisdulici:
Closing due to no activity.

PR #807, opened 2019-03-30
Improve dropdown for custom financial year

sausin:
Currently the dropdown on reports just shows a year which doesn't have much of a meaning when talking about financial years start from a month other than January.

This PR improves the clarity on this subject.


sausin:
A separate PR shortly to change the current behavior when change of calendar year automatically makes the defaults jump to the subsequent financial year (which is correct behavior only when financial year starts in January)

denisdulici:
Closed due to out of date.

PR #831, opened 2019-05-28
Include links to bills on relevant transactions

sausin:
In the current setup, the transaction page lists the amounts and other details. However, when there are many bills on an account, it's not straightforward to identify which transaction corresponds to which bill (especially when payments have been performed in multiple transactions)

This PR adds hyperlinks to the amount to ease the identification of the relevant bill.


denisdulici:
Please, resubmit this according to 2.0 version.

Issue #830, opened 2019-05-25
Dashboard cashflow is not correct when financial year != calendar year


baradhili:
Please, submit only real issues. Use the forum for support, feature requests, proposals, new versions, help etc. https://akaunting.com/forum

### Steps to reproduce the issue
Set financial year start to 1 july in settings
Ensure there are transactions recorded in the current financial year

### Expected result
cashflow should show time period of current financial year in top right, cashflow line should reflect entries

### Actual result
cashflow line reflects next financial year, and thus does not reflect entries. Changing cashflow to "previous finaincial year" shows correct timeline

### System information (Akaunting, PHP versions)
current Akaunting, current php 

### Additional comments
Bug is in /app/Traits/DateTime/getFinancialStart

quick n dirty fix is 
```
    public function getFinancialStart()
    {
        $now = Date::now()->startOfYear();
        
        $setting = explode('-', setting('general.financial_start'));

        $day = !empty($setting[0]) ? $setting[0] : $now->day;
        $month = !empty($setting[1]) ? $setting[1] : $now->month;
        $year = $now->year - 1; //this will bring the current financial start to actual year start, but may not work when FY = Calendar year

        return Date::create($year, $month, $day);
    }
```

PR #840, opened 2019-06-12
Fixes Issue #830

baradhili:
Changed calculation in Traits/DateTime/getFinancialStart so that it correctly calculates FY start regardless if FY == calendar year or not


baradhili:
Thats weird.. travis fails with a test for Bills... but no useful details :(

denisdulici:
This should be fixed in 2.0 version.

PR #858, opened 2019-08-17
Fix create response in invoice payments API

JonnyFunFun:
Hi!  While working with the Akaunting API (integrating it with some various other systems we've got in place), I noticed a small bug when creating invoice payments via the API.  The URL passed in the Location: header was missing a slash.  I went ahead and added it 😃


denisdulici:
This has been fixed in 2.0 version.

PR #859, opened 2019-08-18
PostgreSQL compatibility

shahalpk:
While trying to integrate with PostgreSQL i've faced few difficulties, which i've corrected here. 

I had raised a PR #797 earlier, but it was closed due to inactivity. Since I want to use a different branch now for merging, i'm creating a new PR here. From the previous PR, the below points from @hmert and @denisdulici are still pending to be resolved, which i'll complete soon.

1. Installation (from console) is breaking with the new changes. - I have to check and review the issue here.
2. Adding Database Prefix in the installer - Resolution yet to be finalized. Waiting for input from @hmert on how to proceed.


denisdulici:
@shahalpk Can't accept as it breaks the console.

Issue #873, opened 2019-09-13
Issue with print Invoices - error Image::make


lucianobapo:
### Steps to reproduce the issue
When we try to open an invoice print without logo, it call the fallback option of default logo on line 39 of file app/Http/ViewComposers/Logo.php
$path = asset('public/img/company.png');

### Expected result
Expected to see invoice print

### Actual result
500 Error on Image::make
cannot read the complete url of default file

### System information (Akaunting, PHP versions)
PHP 7.2.19-0ubuntu0.19.04.2 (cli) (built: Aug 13 2019 11:45:23) ( NTS )
Akaunting 1.3.17

### Additional comments
When we change the code to: 
file app/Http/ViewComposers/Logo.php
39: $path = 'public/img/company.png';

It works fine!

PR #874, opened 2019-09-13
Issue with print Invoices

lucianobapo:
Issue with print Invoices - error Image::make #873
When we try to open an invoice print without logo, it call the fallback option of default logo on line 39 of file app/Http/ViewComposers/Logo.php
$path = asset('public/img/company.png');
500 Error on Image::make
cannot read the complete url of default file

When we change the code to:
file app/Http/ViewComposers/Logo.php
39: $path = 'public/img/company.png';

It works fine!

### DIFF
--- a/app/Http/ViewComposers/Logo.php
+++ b/app/Http/ViewComposers/Logo.php
@@ -36,7 +36,7 @@ class Logo
                 return $logo;
             }
         } else {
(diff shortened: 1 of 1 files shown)


denisdulici:
This should be fixed in 2.0 version.

PR #875, opened 2019-09-13
Code Improvement - Select Invoice Items with 0 quantity - Items.php

lucianobapo:
When we remove this condition we can select items with 0 stock quantity on Invoices.
Suggest to make it as an option in settings.


xiris:
@lucianobapo would be better if you send the PR with the suggested change. (add as an option into settings)

denisdulici:
As Christopher said, feel free to resubmit the PR with settings.

PR #876, opened 2019-09-13
Code Improvement - Dynamic conversion

lucianobapo:
Put code at Line 53: if ($default==$code) return $amount;
Then function will not convert amount with same currency codes.


lucianobapo:
updated to #877

PR #878, opened 2019-09-13
Code Improvement - Conversions

lucianobapo:
Put code: "if ($default==$code) return $amount;"

Then function will not convert amount with same currency codes.


denisdulici:
This has been added to 2.0 version.

PR #885, opened 2019-10-13
Improve current Docker files

josepostiga:
While inspecting this project, I noticed that the Docker files were missing a little love. So, I took the liberty to do a little refactoring to them:

1. Moved all build/infrastructure related files to a dedicated `.docker` folder, leaving only the `docker-compose.yml` file in the root of the project.
2. Completely overhaul the PHP docker image, removing the dependency of the `php:apache` base image for a more isolated one, `php:7.2-fpm`. This container has the required dependencies pre-installed (like `pdo_mysql`, for example) and has also xDebug installed and activated (for easier debug sessions).
3. Since the apache server was stripped out, I added a dedicated NGINX container to handle the server requests, that proxies to the PHP container. This is pre-configured to handle requests to the domain `akaunting.test`.

I think that with this refactor, the infrastructure-related files and the actual Akaunting software files don't get mixed in the root folder.

I hope to be of any help! :+1:


denisdulici:
Jose, we've moved docker files to their own repo, feel free to contribute there https://github.com/akaunting/docker

PR #942, opened 2019-11-27
Missing bill payment relationship

mj-isip23:


mj-isip23:
I'm not sure but it looks like the relationship with Bill and Payment is not defined causing an error when showing a bill.

I added some additional model configuration like traits, fillables, and some scope which I based on Transaction model

PR #943, opened 2019-11-27
Fixed bill show page

denisdulici:
close #942


mj-isip23:
@denisdulici you forgot the one on the Bills controller on prepareBill method

PR #954, opened 2019-11-28
Go to dashboard

spinzar:
This button does not translate well as it can change depending on the variable.

### DIFF
--- a/resources/lang/en-GB/general.php
+++ b/resources/lang/en-GB/general.php
@@ -122,7 +122,7 @@ return [
     'disable'               => 'Disable',
     'select_all'            => 'Select All',
     'unselect_all'          => 'Unselect All',
-    'go_to'                 => 'Go to :name',
+    'go_to'                 => 'Go to dashboard',
     'created_date'          => 'Created Date',
     'period'                => 'Period',
     'frequency'             => 'Frequency',


denisdulici:
Can you please elaborate on the issue?

spinzar:
please chack the URL link
http://localhost/akaunting/wizard/finish

in tr-TR general.php line 45    'dashboard'  => 'Kontrol Paneli',
in en-GB general.php line 52    'dashboard'  => 'Dashboard',
in js-JP general.php line 52    'dashboard'  => 'ダッシュ ボード',

after installing in local js-JP
there is something strange, chacke the URL link http://localhost/akaunting/wizard/finish 
Control panel button instead of Go to Dashboard it links  to 
general.php line 125 'go_to'    => 'に行く：名前',
instead of general.php line 'dashboard'   => 'ダッシュ ボード', which we need !
when i amended  general.php line 125 'go_to'    => 'に行く：名前',  as below
'go_to'                 => 'Go to :name',
'go_to'                 => 'Go to Dashboard,
then button error disappeared

spinzar:
japanese

spinzar:
after change

denisdulici:
That is because you've translated the `:name` string which shouldn't as it's a dynamic parameter and that has been done to the whole Japanese translation. Can you please fix all those translations? It should be like this:

`My translation :name`

More info: https://laravel.com/docs/6.x/localization#replacing-parameters-in-translation-strings

spinzar:
Thanks Denis Duliçi
OK i got it.

denisdulici:
Please, let me know when you've finished the review of Japanese translation so I could update it.

PR #955, opened 2019-11-29
Bulk action class discovery for modules

mj-isip23:
Current way of discovering bulk action class for modules is using the view.

`Blade: moduleName::group.blade` will search for a bulk action class `Module\
moduleName\BulkActions\moduleName::group`

### DIFF
--- a/app/Http/ViewComposers/Index.php
+++ b/app/Http/ViewComposers/Index.php
@@ -38,7 +38,7 @@ class Index
         if (Str::contains($view_name, '::')) {
             $names = explode('::', $view_name);
 
-            $params = explode('.', $view_name);
+            $params = explode('.', $names[1]);
 
             $type = $params[0];
 


denisdulici:
I'm not sure I got the issue, could you plz give an example in details?

mj-isip23:
Sorry for giving not enough info. I realized that this is depends on the implementation made by the developer.

Basically I created a model and a blade to show all the data. My file structure for the blade is like this:

* modelPlural
-> index.blade.php

So accessing the blade from the controller will be like this: `moduleName::modelPlural.blade`. The current code will look for a class named: `moduleName::modelPlural`. 

My solution for this is just to use the explode function on the `name` variable instead of the original variable name. I hope it makes sense, I'm open for a workaround for this.

denisdulici:
#979 fixed this issue.

PR #973, opened 2019-12-05
enable/disable mail protocol

mj-isip23:
The inputs for SMTP are not enabled/disabled when the protocol is changed. This is a bug on two machines that I've installed Akaunting. They are only being enabled/disabled when the user clicks on the save button or viewed a different browser page. It is not the best solution but it works


mehmetcoban:
@mj-isip23  Thanks for your solution

PR #985, opened 2019-12-10
Middleware updated in order to decide that wizard is completed or not

SevanNerse:
Middleware updated in order to decide that wizard is completed or not


SevanNerse:
In order to reorganize pull request

Issue #934, opened 2019-11-22
Installation: Cannot Delete Default Currencies


rdrane:
Please, submit only real issues. Use the forum for support, feature requests, proposals, new versions, help etc. https://akaunting.com/forum

### Steps to reproduce the issue
Create a new installation of Akaunting 2.0.1.
Follow the installer wizard and connect to a new (empty) database.
Continue the wizard until you get to the "Currencies" page. 
Try to delete any of the default currencies. 

### Expected result
The currency is deleted.

### Actual result
The screen hangs on the popup and you cannot remove the currency. 
The browser console displays: 
`DELETE http://localhost/Akaunting_2.0.1-Stable/index.php/wizard/currencies/2 405 (Method Not Allowed)`
`Error: Request failed with status code 405`
`    at createError (currencies.js?v=2.0.1:2200)`
`    at settle (currencies.js?v=2.0.1:2363)`
`    at XMLHttpRequest.handleLoad (currencies.js?v=2.0.1:1733)`

### System information (Akaunting, PHP versions)
Akaunting 2.0.1
PHP 7.3.11

PR #989, opened 2019-12-11
Deleting problem of currencies at wizard step is fixed. It also solve…

SevanNerse:
…s issue #934

Feature tests are added.

### DIFF
--- a/database/seeds/Roles.php
+++ b/database/seeds/Roles.php
@@ -17,12 +17,12 @@ class Roles extends Seeder
     public function run()
     {
         Model::unguard();
-
+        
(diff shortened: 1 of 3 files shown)


database/seeds/Roles.php
```diff
@@ -139,41 +139,41 @@ private function map()
             'u' => 'update',
             'd' => 'delete'
         ];
-
+
```
AhtishamShahid: Please Remove empty spaces and formatting changes in all files.

SevanNerse: Thanks for your assistance

PR #998, opened 2019-12-14
Radio group partial is fixed.

SevanNerse:
Radio group controls are not working properly.


ysfkaya:
The setting properties does not working. Laravel throws exception when call the setting property. Try to calls property with "get" method. Then it work well.

Ekran Resmi 2019-12-14 11 18 09

SevanNerse:
Thanks for your kindly assistance. @ysfkaya

Issue #982, opened 2019-12-10
Modals aren't working


mj-isip23:
Please, submit only real issues. Use the forum for support, feature requests, proposals, new versions, help etc. https://akaunting.com/forum

### Steps to reproduce the issue
1. Navigate to items
2. Create new item
3. On tax select, click 'add new tax'

### Expected result
I'm pretty sure that a modal should pop up cause that is the UX for 1.3.17 version.

### Actual result
Nothing happens.

### System information (Akaunting, PHP versions)
Akaunting 2.1
Php 7.2.12

### Additional comments
I'm not sure if this is an issue or the devs are just not working on it yet.

PR #1011, opened 2019-12-19
fix add tax modal

mj-isip23:


mj-isip23:
Fixes issue #982

cuneytsenturk:
Thanks Michael John

PR #1021, opened 2019-12-21
Error while running `php artisan install ...` if mysql/mariadb has `d…

alfredleo:
Error while running `php artisan install ...` if mysql/mariadb has `default-storage-engine=MYISAM` instead of `InnoDB`. This will ensure that you won't get this error in any case.

Fix to SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was too long; max key length is 1000 bytes

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -57,7 +57,7 @@ return [
             'prefix' => env('DB_PREFIX', 'ak_'),
             'prefix_indexes' => true,
             'strict' => true,
-            'engine' => null,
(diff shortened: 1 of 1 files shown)


alfredleo:
akauntingError

denisdulici:
Alfred, what is the RAW_FORMAT for?

alfredleo:
If you just write InnoDB, mysql deaults to COMPACT.
The best option for CPU usages but not space is Dynamic.
INNODB

PR #1024, opened 2019-12-23
Fixing style of description column on transactions page. It should wr…

SevanNerse:
…ap in order to handle long descriptions.
image


cuneytsenturk:
Thank your commit.
Please change `text-wrap` to `long-texts`. 
We use this class for all system.

PR #1026, opened 2019-12-24
Value of radio group controls should be given on edit page of settings

SevanNerse:
System have to handle value for radio group controls.


cuneytsenturk:
Thanks @SevanNerse

PR #1032, opened 2019-12-25
Bulgarian usage is like this: 1 234 567,89

ilhansubasi:
Bulgarian usage is like this: 1 234 567,89 Also see https://en.wikipedia.org/wiki/Decimal_separator


denisdulici:
Thanks İlhan

PR #1043, opened 2019-12-27
Allow different field sizes across modules & more

menezesbr:
Allow different field sizes across modules and remove icon space when not defined
exemple setting from module:
"settings": [
        {
            "type": "textGroup",
            "name": "username",
            "title": "nfsebru::general.username",
            "icon": "",
            "attributes": {
                "required": "required",
                "col": "col-md-3"
            }
        },


denisdulici:
You don't need to use `Form::textGroup` if you're not going to use HTML group, feel free to use `Form::text` directly.

PR #1149, opened 2020-01-21
Human friendly import/export

denisdulici:


app/Traits/Import.php
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    public function getAccountIdFromCurrency($row)
+    {
+        return Account::firstOrCreate([
+            'currency_code'     => $row['currency_code'],
+        ], [
+            'company_id'        => session('company_id'),
+            'name'              => !empty($row['account_name']) ? $row['account_name'] : $row['currency_code'],
+            'number'            => !empty($row['account_number']) ? $row['account_number'] : Account::max('number') + 1,
```
burakcakirel: 'number' field is not numeric. It's a string field:
https://github.com/akaunting/akaunting/blob/2.0.1/database/migrations/2017_09_01_000000_create_accounts_table.php#L19
https://github.com/akaunting/akaunting/blob/2.0.1/app/Http/Requests/Banking/Account.php#L28

So, `Account::max('number') + 1` section will raise an error

app/Traits/Import.php
```diff
@@ -0,0 +1,212 @@
...
+        ])->id;
+    }
+
+    public function getAccountIdFromName($row)
+    {
+        return Account::firstOrCreate([
+            'name'              => $row['account_name'],
+        ], [
+            'company_id'        => session('company_id'),
+            'number'            => !empty($row['account_number']) ? $row['account_number'] : Account::max('number') + 1,
```
burakcakirel: Same here:
https://github.com/akaunting/akaunting/pull/1149/files#r369201551

app/Traits/Import.php
denisdulici: You're right, what do you suggest?

burakcakirel: I think we need to add a rule like the following one for the 'account_number' field:
https://github.com/akaunting/akaunting/pull/1148/files#diff-140f210bef6fa64d899a53cb3996e441R105

denisdulici: Or maybe something like this https://paulund.co.uk/increment-numeric-part-of-string

PR #1158, opened 2020-01-21
Sample data seeder

burakcakirel:


database/seeds/SampleData.php
```diff
@@ -0,0 +1,109 @@
...
+     */
+    public function run()
+    {
+        Model::reguard();
+
+        $faker = Factory::create();
+
+        $count = $this->command->option('count');
+
+        for ($i = 0; $i < $count; $i++) {
```
denisdulici: Why not use factory directly like `factory(Contact::class, $count)->create();` ? It will persist the db.

database/seeds/SampleData.php
```diff
@@ -0,0 +1,109 @@
...
+            $this->dispatch(new CreateAccount(factory(Account::class)->raw()));
+        }
+
+        for ($i = 0; $i < $count; $i++) {
+            $this->dispatch(new CreateBill(factory(Bill::class)->state('items')->raw()));
+        }
+
+        for ($i = 0; $i < $count; $i++) {
```
denisdulici: Same as above, factory should be enough `factory(Invoice::class, $count)->create();`

database/seeds/SampleData.php
burakcakirel: I thought that we need all processes in jobs like creating items and totals for invoices but it's already done by factories

database/seeds/SampleData.php
```diff
@@ -0,0 +1,95 @@
...
+
+        $count = $this->command->option('count');
+
+        factory(Contact::class, (int)$count)->create();
+        factory(Item::class, (int)$count)->create();
+        factory(Account::class, (int)$count)->create();
+        factory(Bill::class, (int)$count)->create();
+        factory(Invoice::class, (int)$count)->create();
+
+        for ($i = 0; $i < $count; $i++) {
```
denisdulici: This is not required as the invoice/bill factory will do it for paid documents.

database/seeds/SampleData.php
denisdulici: Jobs are required for real application but not for dummy data 😉

denisdulici: BTW, factories use jobs in [some](https://github.com/akaunting/akaunting/blob/master/database/factories/Invoice.php#L127) cases.

PR #1161, opened 2020-01-21
Show grand total for bills/invoices

burakcakirel:
Show grand total for bills/invoices in "Amount" field in "Add Payment" modal

### DIFF
--- a/app/Http/Controllers/Purchases/Bills.php
+++ b/app/Http/Controllers/Purchases/Bills.php
@@ -75,8 +75,6 @@ class Bills extends Controller
 
         $date_format = $this->getCompanyDateFormat();
 
-        $paid = $this->getPaidAmount($bill);
-
         // Get Bill Totals
         foreach ($bill->totals as $bill_total) {
             $bill->{$bill_total->code} = $bill_total->amount;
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Purchases/Bills.php
```diff
@@ -74,6 +75,21 @@ public function show(Bill $bill)
 
         $date_format = $this->getCompanyDateFormat();
 
+        $paid = $this->getPaidAmount($bill);
```
denisdulici: Why not use the [paid](https://github.com/akaunting/akaunting/blob/master/app/Models/Purchase/Bill.php#L209) attribute of model?

app/Http/Controllers/Sales/Invoices.php
```diff
@@ -77,6 +78,21 @@ public function show(Invoice $invoice)
 
         $date_format = $this->getCompanyDateFormat();
 
+        $paid = $this->getPaidAmount($invoice);
```
denisdulici: Same [here](https://github.com/akaunting/akaunting/blob/master/app/Models/Sale/Invoice.php#L218).

PR #1225, opened 2020-02-03
Remove fontawesome depency in form fields

SevanNerse:
…, components/AkauntingSearch.vue, components/AkauntingSelectRemote.vue) and related forms


denisdulici:
While I support this change, it must be applied to all form fields, not only select. Furthermore, all modules must be updated so we should prepare this for 2.1 version I guess.

PR #1265, opened 2020-02-17
Fix amount precision in Invoice and Bill factory

burakcakirel:


database/factories/Invoice.php
```diff
@@ -144,7 +144,7 @@
             ];
 
             if ($init_status == 'partial') {
-                $payment_request['amount'] = (double) $amount / 3;
+                $payment_request['amount'] = round($amount / 3, $invoice->currency->precision);
```
hmert: Rounding may cause errors. @denisdulici

denisdulici: It's just sample data, nothing to worry about 😉

PR #1267, opened 2020-02-18
multiple prop of AkauntingSelect.vue component should be passed as a …

SevanNerse:
…dynamic prop on multi_select_add_new_group.blade.php

It throws an exception like below.

image


cuneytsenturk:
Thanks Sevan

PR #1366, opened 2020-03-18
Add item & contact CRUD notification events

demirhanaydin:
When an item or contact is created, deleted, or updated, an event will be fired

### DIFF
--- a/app/Events/Common/ContactCreated.php
+++ b/app/Events/Common/ContactCreated.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace App\Events\Common;
+
+use Illuminate\Queue\SerializesModels;
+
+class ContactCreated
+{
+    use SerializesModels;
+
+    public $contact;
+
+    /**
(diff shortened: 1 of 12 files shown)


denisdulici:
Any reason for not using Eloquent events? https://laravel.com/docs/eloquent#events

denisdulici:
You can use Observers to listen to Eloquent events: https://laravel.com/docs/eloquent#observers

demirhanaydin:
Thanks a lot, I'll check it. Let's ignore this pr for now.

denisdulici:
You're welcome 😉  

We generally fire events manually if there is more than Eloquent CRUD i.e. [invoice creation](https://github.com/akaunting/akaunting/blob/master/app/Jobs/Sale/CreateInvoice.php#L59).

denisdulici:
[Here](https://akaunting.com/docs/developer-manual/hooking-models) I just created an example.

PR #1372, opened 2020-03-21
Discount per Item

burakcakirel:


database/migrations/2020_01_08_000000_core_v200.php
```diff
@@ -240,6 +240,11 @@ public function up()
         Schema::table('users', function (Blueprint $table) {
             $table->string('landing_page', 70)->nullable()->default('dashboard');
         });
+
+        Schema::table('invoice_items', function (Blueprint $table) {
+            $table->double('discount_rate', 15, 4)->default('0.0000');
+            $table->string('discount_type')->default('normal');
+        });
```
denisdulici: this must be into a separate migration file

app/Http/Controllers/Settings/Defaults.php
```diff
@@ -27,12 +27,18 @@ public function edit()
 
         $payment_methods = Modules::getPaymentMethods();
 
+        $discount_locations = [
+            'in_totals' => trans('settings.default.discount_in_totals'),
+            'per_item' => trans('settings.default.discount_per_item'),
+        ];
+
```
denisdulici: Redundant

app/Http/Controllers/Settings/Defaults.php
```diff
@@ -27,12 +27,18 @@ public function edit()
...
         return view('settings.default.edit', compact(
             'setting',
             'accounts',
             'currencies',
             'taxes',
-            'payment_methods'
+            'payment_methods',
+            'discount_locations'
```
denisdulici: also this one

PR #1375, opened 2020-03-22
Don't overwrite the bg-variant mixin

blackpidgeon:
Bootstrap 4.3 introduced an additional parameter, and npm run dev fails
because we are overwriting a mixin that bootstrap uses internally.

Also update the vue2-transition module to the same version of the dev..


cuneytsenturk:
Thank you @blackpidgeon

PR #1403, opened 2020-04-10
added bill/invoice number to pdf filenames

mocig:
Added the invoice number to the filename:
From invoice_timestamp.pdf
To: invoice_invoicenumber_timestamp.pdf

Timestamp kept in filename because invoices can be edited and therefore there diffrent versions over time possible.

### DIFF
--- a/app/Http/Controllers/Purchases/Bills.php
+++ b/app/Http/Controllers/Purchases/Bills.php
@@ -342,7 +342,8 @@ class Bills extends Controller
 
         $pdf = app('dompdf.wrapper');
         $pdf->loadHTML($html);
-	$safe_bill_number = preg_replace("/[^A-Za-z0-9_-]/", '', $bill->bill_number);
+
+        $safe_bill_number = preg_replace("/[^A-Za-z0-9_-]/", '', $bill->bill_number);
         $file_name = 'bill_'.$bill->bill_number.'_'.time() . '.pdf';
 
         return $pdf->download($file_name);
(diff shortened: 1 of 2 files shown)


denisdulici:
What happens when the invoice/bill number contains special chars?

mocig:
> What happens when the invoice/bill number contains special chars?

Very good point. Easiest solution strip all forbidden from bill/invoice number. Add that in case you agree. Worst case all are invalid and results in old filename but with __ instead of _.

denisdulici:
Regex makes sense only for Latin alphabet. For others, it will even make worse having strange numbers in file name 😞

mocig:
> Regex makes sense only for Latin alphabet. For others, it will even make worse having strange numbers in file name 😞

I was thinking about using \p{u} to allow unicode letters, but what if the underlaying filesystem doesn't supports it? Having a timestamp in the filename like invoice_1586703459.pdf ALREADY is a filename with strange numbers. Not much diffrent that invoice_13_1586703459 .pdf. BOTH is strange numbers in filename. But at least user has the _ability_ to use something usefull.

Well, it's your project. I'll apply it to my local instance because I def. agree that this is a usefull change , and I added it as pull request because I think for _most_ users it is aswell. For all other users theres no big diffrence, invoice_numbers_numbers.pdf or invoice_numbers.pdf looks same way to me.

Your call brother :)

mocig:
Btw, filename INVOICE_ and BILL_ is latin alphabet anyway. Orgininal source doesn't names it 發票_timestamp.pdf neigher.

denisdulici:
I preferred to use `Str::slug` as it transliterates UTF-8 to ASCII.

mocig:
> I preferred to use `Str::slug` as it transliterates UTF-8 to ASCII.

Very nice. 1+

PR #1405, opened 2020-04-12
Update passwords.php

tfevan:
Added missing key and value


denisdulici:
Please, use Crowdin for translations https://crowdin.com/project/akaunting

PR #1406, opened 2020-04-13
Sort invoice totals by sort_order column

PKeidel:
List of totals was not sorted correctly.


denisdulici:
This should not be in the model but controller. Also, can you please apply to bills too?

PR #1427, opened 2020-04-26
German: deleted "," after valediction

characta:
This is done in English but in German one does not write a comma after the last closing words before a name.
See here:
https://de.wikipedia.org/wiki/Gru%C3%9Fformel_(Korrespondenz)


denisdulici:
Can you please use Crowdin for language changes? https://crowdin.com/project/akaunting

PR #1429, opened 2020-04-26
Use config helper instead of env in application code

braceyourself:
Based on the warning in laravel's docs: [configuration-caching](https://laravel.com/docs/7.x/configuration#configuration-caching)

This PR simply replaces calls to env() with config(), allowing the configuration to be cached, improving app performance.

Also, added a call to 'artisan config:cache' after calling Installer::updateEnv.

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -23,11 +23,11 @@ class Kernel extends ConsoleKernel
     protected function schedule(Schedule $schedule)
     {
         // Not installed yet
-        if (!env('APP_INSTALLED')) {
+        if (!config('app.installed')) {
             return;
         }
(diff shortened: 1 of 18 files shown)


denisdulici:
Thanks @braceyourself for the huge work.

We [override](https://github.com/akaunting/akaunting/blob/master/app/Utilities/Overrider.php#L33) some configs so have you tested config caching with them?

braceyourself:
I looked into getting the overridden settings cached properly, but it will take quite a bit of work - more than just a simple find and replace. ;)

denisdulici:
Replacing env usage with config does also matter so going to merge it and remove caching.

PR #1432, opened 2020-04-29
Updated default start date to have a 1 day and month.  

BrianLakstins:
0 is not a valid day or a month.
Using 0000-00-00 with MySQL 8.0.20 causes error SQLSTATE[HY000]: General error: 1525 Incorrect DATE value


denisdulici:
I'm not able to reproduce such an issue, see screenshot. Also, the end of the month is not correct too.

Screenshot (97)

BrianLakstins:
The error occurs for me when either date is 0000-00-00.  It does not occur when the date is 0000-01-01.  It occurs when I attempt to create a new reconciliation.

image

BrianLakstins:
Nevermind.  I see you added a better solution with this commit.

https://github.com/akaunting/akaunting/commit/998983ca0fccaa93f456a96510b557db0c9c85bc

PR #1465, opened 2020-06-04
Show dynamic component also in portal

burakcakirel:


cuneytsenturk:
@burakcakirel Thank you.

PR #1504, opened 2020-06-25
allow using own already selected checkbox values

chistel:
currently the checkbox partial blade does not allow you set own selected values, which prevents some dynamic options based on wheat you intend to work on.

this pull request is backward compatible with every play where the checkboxGroup macro is used


denisdulici:
Shouldn't the forms that already use checkbox field be updated too?

cuneytsenturk:
Thank you @chistel for your contribution.

PR #1535, opened 2020-07-15
Update item total using the item tax total upon changing taxes in invoice detail

boumanb:
There are quite some errors here. Sometimes the tax doesn't get calculated and other times the currency sign is being omitted. It's probably due to combined tax types. They give strange results when playing with them, without any notification to the user. Could make a PoC video.

This tiny modification makes it somewhat more stable for me at least.


cuneytsenturk:
Hi Billy Bouman,

I've check this PR but we can not merge it. Because Invoice item row not show tax amount.
it show total groups.

Thank you for your contribution.

Have a good day

Issue #1551, opened 2020-07-27
Reconciliation form checkbox missing values


spronych:
See #1526 for reference and steps to reproduce.

### System information (Akaunting, PHP versions)

Akaunting 2.0.19, PHP 7.4.6

### Additional comments

This error is happening because there is an unknown value specified in the Form::checkbox function on line 86 of the /resources/views/banking/reconciliations/create.blade.php file. The variable "$item->amount_for_account" should actually be "$item->amount"? The change originally happened with 630fdb6.

Corrected code is as follows:

`Line 86:  {{ Form::checkbox($item->type . '_' . $item->id, $item->amount, $item->reconciled, [`

spronych:
The file /resources/views/banking/reconciliations/edit.blade.php also needs to have line 58 corrected to the following:

`{{ Form::checkbox($item->type . '_' . $item->id, $item->amount, $item->reconciled, [`

PR #1556, opened 2020-08-03
Reconciliation form checkbox missing values #1551

spronych:


cuneytsenturk:
This commit not working. it should be `$item->amount_for_account`. Because Akaunting support multi currency

Issue #1346, opened 2020-03-10
Search/Filter not working


aliuysal60:
### Steps to reproduce the issue

Open Invoices Lists
search any word

### Actual result

no data

### System information (Akaunting, PHP versions)
2.0.4
PHP 7.4

### Additional comments

Invoices and Purchases, Customers, Vendors. The search box in their list is not working

![image](https://user-images.githubusercontent.com/6507261/76346825-b7a9a480-6316-11ea-976b-13b73d25719c.png)

innovusvm:
Same problem here! Also in previous versions was very helpful the hability to filter by customer and status.

alex-phillips:
Also filtering by date range in the previous version was quite useful

dungeonist:
Same here. I can't filter data in invoice and expenses area. Any word or information return a "No Data" as in the screenshot of the bug report.

alanbrande:
Not working for me also, would be great to be able to filter per column as it was before.

simkea:
Is anyone currently working on this problem?
I can confirm, the internal search in the listings does not work.
There is no query/ajaxcall to the server.

fatihalp:
Most basic filter is not working for months even on cloud. I believe it's urgent and blocker issue @cuneytsenturk

alex-phillips:
Agreed. This is the sole reason I still haven't upgraded to v2...

hexmarkrecords:
This feature is core to the accounting platform. We need reports based on a period to file taxes.

aletfox:
anyone know the status of this bug?

domaniEquestrian:
We are facing the same issue..

fatihalp:
Isn't it most basic need @denisdulici ? 

How I'll find my customer in 10.000 customer? Shoud be urgent?

vrielsa:
I'm working on a PR for this. 
I'll try to finish it tonight.

hexmarkrecords:
When fixed do you know how long it takes to make it to production?

regedittap:
Hi any update on this bug?

3run01:
Oi alguma atualização sobre este bug?

3run01:
Hi any update on this bug?

mehdi89:
Any update here? it seems updating to the latest version was a mistake. I don't understand why such an important feature was missing in the first place. 

The basic filter itself and Filter by date and each column.

jleu84:
[...]

PR #1563, opened 2020-08-08
Made items, Purchases and Sales searchable #1346

vrielsa:
Hey

I've added search capabilities to the following pages:
* Items
* Bills
* Payments
* Vendors
* Customers
* Invoices
* Revenues

I've written 'SearchResultCollectors' as separate services in order to have thin controller and to make them reuse-able. Each provider implements the SearchCollecterInterface. In a later state we could tag the collecters with a service provider and inject them in the common/search controller (just a wild thought).

Functionality wise, when typing in the search bar it uses the function 'usingSearchString' in the backend. So you could actually search with: `date:2020` or something or with `paid_at:2020`. When selecting a search option you get redirected to the selected entities show or in some cases the edit page (revenue and payments).

Addresses issue #1346

### DIFF
--- a/app/Http/Controllers/Common/Items.php
+++ b/app/Http/Controllers/Common/Items.php
@@ -15,7 +15,9 @@ use App\Models\Common\Item;
 use App\Models\Setting\Category;
 use App\Models\Setting\Currency;
 use App\Models\Setting\Tax;
+use App\Services\Search\Common\ItemCollector;
 use App\Traits\Uploads;
+use Illuminate\Http\JsonResponse;
 
 class Items extends Controller
 {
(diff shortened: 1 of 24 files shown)


hexmarkrecords:
Hi thanks for your work on this. I used the double entry accounting plugin, does that mean that the developer would need to implement the search based on your PR?

vrielsa:
No problem, with pleasure :-) This just has to be checked by the developers of Akaunting or the maintainers before it can be used. The feature wasn't completely implemented yet, so they might have other idea's for it.

If the Double-Entry app comes with new search fields, the developers of it will have to implement the logic for it indeed.

denisdulici:
First of all, thank you Sarah for the huge work. However, it doesn't solve the problem of missing filters in the UI.

End-users can't write `date:2020` or any kinda search string. Therefore, @cuneytsenturk is working on making a GitLab-like search bar with filters.

PR #1570, opened 2020-08-13
Remove long spacing in e-mails

ctsoft-de:
If you edit the e-mail templates the Quill editor adds `<p>` tags around all the `<br>` tags. In many mail clients this leads to very long spacing. This PR removes the `<p>` tags when rendering the e-mails.

Before:
<kbd>image</kbd>

After:
<kbd>image</kbd>


denisdulici:
Unfortunately, can't merge this as the fix should be applied during the saving process, not show.

PR #1573, opened 2020-08-18
Development environment

asafov:
Development environment with Laravel Homestead and Vagrant with examples.

### DIFF
--- a/Homestead.yaml
+++ b/Homestead.yaml
@@ -1,0 +1,16 @@
+ip: 192.168.10.10
+memory: 2048
+cpus: 2
+provider: virtualbox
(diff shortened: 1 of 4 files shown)


denisdulici:
It seems that PhpStorm is required, right?

Also, I'd prefer having a separate repo for this, like [docker](https://github.com/akaunting/docker), is that possible?

asafov:
PHPStorm don't needed. It's just example where stored source of project.
It can be any. About separate - you want separate functions or config file?
I saw many projects, which include Vagrantfile in repo, so only Homestead.yaml is only for Laravel.

PR #1576, opened 2020-08-23
Remove long spacing in e-mails

domkrm:
See https://github.com/akaunting/akaunting/pull/1570

This PR applies the fix during the saving process.


cuneytsenturk:
I've fixed this [commit](https://github.com/akaunting/akaunting/commit/74a98589d5306e002acb242b0ca34c1075d5be36) so then I am closing this PR

PR #1598, opened 2020-09-10
Add multiple option to select-group

rowino:


cuneytsenturk:
You can use this form element -> https://github.com/akaunting/akaunting/blob/master/resources/views/partials/form/multi_select_group.blade.php for example https://github.com/akaunting/akaunting/blob/master/resources/views/auth/users/create.blade.php#L51

cuneytsenturk:
I'm closing this pull request.

PR #1631, opened 2020-10-07
Fix duplicate stack for item name input in edit invoice view

domkrm:
I wanted to add some custom html after the input for the item name in the edit invoice view. So I used the stack ```name_input_end```. But my custom html showed up after the name input AND after the price input. This is because the ```Form::moneyGroup``` generates a dynamic stack with the passed name argument. And the argument was set to ```name``` but I think it should be ```price```.


cuneytsenturk:
Thanks @domkrm

PR #1636, opened 2020-10-09
Add possibility to set the description per line item in invoices and bills

domkrm:
This PR adds the functionality to set the description per line item in invoices and bills.

For existing invoices and bills the description will be used from the assigned item.


denisdulici:
@cuneytsenturk is working on implementing this in a different way, it'll be available soon.

domkrm:
Any plans when this will be available? I need this urgently and I think of creating a module until then.

PR #1637, opened 2020-10-09
Add page for bill settings

domkrm:
This PR adds a page for bill settings:

- Number prefix
- Number digits
- Next number


denisdulici:
Thanks but this is not really used by most users.

domkrm:
@denisdulici But I think for the users who need this (like me) this will be very helpful, because then you do not have to play around in the database. And I think it doesn't hurt anybody if this settings are available.

PR #1654, opened 2020-11-03
path of ManagesStacks.php is fixed

SevanNerse:
This situation prevents working of some other moduls properly.
It throws below warning message as well.

Ekran görüntüsü 2020-11-03 122007


cuneytsenturk:
Thanks @SevanNerse

PR #1659, opened 2020-11-09
Fix error when change account enabled status from account list

domkrm:
It was not possible to change the account enabled status from the account list (error message: Warning: You are not allowed to disable or change the currency...). This is because the ```UpdateAccount``` job always checked the currency code for changes.

Furthermore it was possible to disable the company default account if there are no transactions. This PR fixes this also.

### DIFF
--- a/app/Jobs/Banking/UpdateAccount.php
+++ b/app/Jobs/Banking/UpdateAccount.php
@@ -52,22 +52,26 @@ class UpdateAccount extends Job
      */
     public function authorize()
     {
+        if ($this->request->has('enabled')) {
+            if (!$this->request->get('enabled') && ($this->account->id == setting('default.account'))) {
+                $relationships[] = strtolower(trans_choice('general.companies', 1));
+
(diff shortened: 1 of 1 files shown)


denisdulici:
Please, provide the steps to reproduce the issue.

domkrm:
Disable default account:
Banking -> Accounts -> Create a new account and set it as default account -> Edit the newly created account -> Set enabled to "no" -> Save -> No error message is displayed, the account is disabled and the company doesn't have a default account anymore. If there were any transactions for this account then an error message is displayed if you try to disable the account. I think the correct behaviour should be that you never can't disable the default account (regardless if there are any transactions)

Disable account in account list:
Banking -> Accounts -> Create a new account (not default) -> Save -> In the list try to disable the newly created account -> An error message is displayed

denisdulici:
`master` is not accepting PR for a while so I've applied this to `2.1-dev` as it was with `DeleteAccount` job.

PR #1660, opened 2020-11-09
Use all income types for reconciliations

domkrm:
I developed a module which uses a custom income type. But in the reconciliations all transactions with this custom income type are displayed and calculated as expense. This PR fixes this problem and respects all income types (not only ```income```).


cuneytsenturk:
Thank you @domkrm

PR #1661, opened 2020-11-09
Fix transaction not visible when paid on the end date of a reconciliation

domkrm:
When creating a reconciliation and a transaction was paid on the selected end date (Y-m-d) of the reconciliation and after the current time (H:i:s) then this transaction was not visible in the list. This PR sets ```23:59:59``` as the end time and all transactions paid on the end date will be loaded.


cuneytsenturk:
Thank you @domkrm

PR #1667, opened 2020-11-15
Set automatic notification per customer

tognit:
With this pull request, you could set the customer whether an automatic email should be sent.


denisdulici:
Thanks for your contribution but we're not going to add such a feature to the core.

PR #1674, opened 2020-11-28
EN-US localization (and one typo in EN-GB)

flufftronix:
Hi, went through the GB-US language files and made some minor changes between US and UK English (color/colour, localization/localisation, etc.) and syntax.

Also noticed a missing word in the lost password php file, which I updated in both EN-US and EN-GB. Was:

> The following language lines are the default lines which match reasons that are given by the password broker for a password update attempt has failed, such as for an invalid token or invalid new password.

Is now:

> The following language lines are the default lines which match reasons that are given by the password broker for a password update attempt **which** has failed, such as for an invalid token or invalid new password.

Thanks!


denisdulici:
Thanks for your contribution. Feel free to use Crowdin for translations:

https://crowdin.com/project/akaunting

PR #1699, opened 2020-12-28
minor bugs fixed

SevanNerse:
- document relationship at document related models is fixed by adding `type` field
- visible-change attribute is added to select related form controls
- function defined in order to add fields on document items


cuneytsenturk:
Thansk @SevanNerse 🎉

PR #1701, opened 2020-12-29
Delete lines that duplicate removing the 'id' column from the migrati…

pavel-mironchik:
…ng data query.


cuneytsenturk:
thanks @pavel-mironchik

PR #1702, opened 2020-12-29
Handle a possible issue with orphaned 'document_item_taxes' while upg…

pavel-mironchik:
…rading to 2.1.0.


cuneytsenturk:
thanks @pavel-mironchik

PR #1706, opened 2020-12-31
Fix a width of a column with a document number.

pavel-mironchik:
Or otherwise, the table doesn't take up all the space.
2020-12-31 13-43-52 Ubuntu Mate 19 10 (Снимок 160)  Работает  - Oracle VM VirtualBox   2


pavel-mironchik:
Though, I'd change the width of the "Customer" column instead, because generally, names are much lengthier than a document numbers.

PR #1709, opened 2021-01-01
Fix a typo.

pavel-mironchik:
This PR just fixes a simple typo which doesn't cause any problems actually.


cuneytsenturk:
thanks @pavel-mironchik

PR #1710, opened 2021-01-01
Leverage translation strings in the bulk actions modal.

pavel-mironchik:
Though the `$text` can contain just a plain string, generally it is a translation string and should be used with the localization helper. Otherwise, the bulk actions modal will show a weird text instead of an item type, translated into the current locale. And actually, the `$text` variable is used with a localization helper in two places of the view:

2021-01-01 17-36-01 Ubuntu Mate 19 10 (Снимок 162)  Работает  - Oracle VM VirtualBox   1


cuneytsenturk:
thanks @pavel-mironchik

PR #1711, opened 2021-01-01
Fix some usages of translation strings.

pavel-mironchik:
There is no such node in translation strings as `general.messages`, but there are appropriate strings under the `documents.messages` node.


cuneytsenturk:
thanks @pavel-mironchik

PR #1715, opened 2021-01-02
Fix a filter name in documents list.

pavel-mironchik:
Without a translation string the filter for the `issued_at` field is shown like this:

2021-01-02 17-55-09 Ubuntu Mate 19 10 (Снимок 168)  Работает  - Oracle VM VirtualBox   2

After applying this fix the filter will be shown correctly, however, we'll need to find a way to name it correspondingly to the document type, but this is a completely different issue:

2021-01-02 18-36-44 Ubuntu Mate 19 10 (Снимок 168)  Работает  - Oracle VM VirtualBox   2


cuneytsenturk:
Thanks @pavel-mironchik

PR #1717, opened 2021-01-02
Fix showing a divider in the document actions menu.

pavel-mironchik:
A divider in the document actions menu should be shown only if there are the following menu items. But now it is shown even when there are no other menu items:

2021-01-02 19-00-27 Ubuntu Mate 19 10 (Снимок 168)  Работает  - Oracle VM VirtualBox   2

This PR fixes such a behavior.


cuneytsenturk:
Thanks @pavel-mironchik

PR #1735, opened 2021-01-07
Added Laravel Sanctum 

stojankukrika:
Added Laravel Sanctum to can use on mobile token applications


denisdulici:
First of all, thanks for your contribution.

Why do we exactly need tokens? How are users expected to create tokens? The API ships with [basic auth](https://github.com/dingo/api/wiki/Authentication) by default and it works perfectly. If we still need tokens, we can use the JWT package supported by the Dingo API package, right? More packages mean more dependency.

stojankukrika:
Thank you @denisdulici 
Mostly I see loot mobile apps that work with tokens so I think that will be OK to add it here too. But Ok. I will try to make it using basic auth as you mention, If I have some issue I will let you know. Thank you for your comments.

stojankukrika:
Hi @denisdulici. I tried to use basic auth, but that does not work well in mobile apps also have a security issue to remember password and send it every time to the server because mobile didn't have sessions like a web.

Also, I tried with tymon/jwt-auth which Dingo API proposes for using JWT. Also, find some issues like generate JWT key which can be an issue for current users, and token expiration refresh, which we didn't have if we use Laravel Sanctum.

Let me know what do you think about it?

denisdulici:
You can save/remember the encoded basic auth header instead of email+password. As for sending every time, that's done with token too, right? The security issue is applicable for tokens too as if I get your token, I could access everything permitted to that token.

stojankukrika:
Thanks to you @denisdulici  I learned how basic auth works! Till today I always use JWT authentications and didn't even look is there any other way to do authentication using mobile app and web (laravel) app. Thank you a lot to motivate me to learn it! Soon will be the first version of the mobile app. Do you think it will be better to use the Akaunting play account to publish it or to use my company account?

denisdulici:
Glad to hear that @stojankukrika. Feel free to contact me directly on https://chat.akaunting.com so we could discuss the distribution 😉

PR #1737, opened 2021-01-07
Add Livewire

cuneytsenturk:
Site: https://laravel-livewire.com
Repo: https://github.com/livewire/livewire


config/livewire.php
```diff
@@ -0,0 +1,100 @@
...
+    |
+    | This value sets the path to Livewire JavaScript assets, for cases where
+    | your app's domain root is not the correct path. By default, Livewire
+    | will load its JavaScript assets from the app's "relative root".
+    |
+    | Examples: "/assets", "myurl.com/app".
+    |
+    */
+
+    'asset_url'  => asset('./public/'),
```
pavel-mironchik, line 45: Use the absolute path `/public` instead of a relative one. Otherwise, Livewire generates such a path to its script when you're on the Bills page for example: `/purchases/public/vendor/livewire/livewire.js?id=936e5d0fb0b76b631ba7`.

cuneytsenturk: I've fixed it.

PR #1739, opened 2021-01-08
option_sortable prop is activated for select components

SevanNerse:


SevanNerse:
Thanks @pavel-mironchik 👍

PR #1741, opened 2021-01-09
naming of variables updated and redundancy created by use of the comp…

SevanNerse:
…uted property is changed in sortBy method


cuneytsenturk:
thanks @SevanNerse and @pavel-mironchik

PR #1743, opened 2021-01-11
Remove nullable property from due_at column

burakcakirel:


pavel-mironchik:
@burakcakirel are you sure about this? For example, credit and debit notes don't have the due date, so this field could contain `null` in such cases.

burakcakirel:
Actually, this was `not null` column before. I don't have any idea why I've marked it as `null` :)

However, currently, `due_at` field is required at the frontend for invoices and bills. I thought it should be `not null`.

But I understand your concern about the modules that don't have this field. 

Maybe, sending `due_at` field as a hidden input works for you to skip the form validation or @cuneytsenturk has a better idea about it?

pavel-mironchik:
> Actually, this was `not null` column before. I don't have any idea why I've marked it as `null` :)

Probably, exactly because of credit/debit notes, or otherwise you couldn't be able to migrate them to documents? I'm wondering now will update scripts work after this change or it will require changes too?

Yeah, I could use the hidden field but that requires to put some random date in it to pass validation, right?

pavel-mironchik:
Ok, this will do in my case: 
`{{ Form::hidden('due_at', '1970-01-01 00:00:00', ['id' => 'due_at', 'v-model' => 'form.due_at']) }}`

burakcakirel:
>Probably, exactly because of credit/debit notes, or otherwise you couldn't be able to migrate them to documents? I'm wondering now will update scripts work after this change or it will require changes too?

That's right. Now I have an idea why I've marked it as `null` :smile: 

So, yes, the migration script needs to be changed and I'll open a PR about it.

cuneytsenturk:
> Actually, this was `not null` column before. I don't have any idea why I've marked it as `null` :)
> 
> However, currently, `due_at` field is required at the frontend for invoices and bills. I thought it should be `not null`.
> 
> But I understand your concern about the modules that don't have this field.
> 
> Maybe, sending `due_at` field as a hidden input works for you to skip the form validation or @cuneytsenturk has a better idea about it?

I've set hidden fiel for hide due_at field..

PR #1750, opened 2021-01-12
Allow to use the script component with modules' custom scripts.

pavel-mironchik:
It automatically uses the correct version (core's or module's) to bypass a browser's cache.

An example of usage:

    <x-documents.script
        type="debit-note"
        script-file="modules/CreditDebitNotes/Resources/assets/js/debit_notes/documents.min.js"
        :items="$debit_note->items"
    />

### DIFF
--- a/app/View/Components/Documents/Script.php
+++ b/app/View/Components/Documents/Script.php
@@ -15,6 +15,9 @@ class Script extends Component
     /** @var string */
     public $scriptFile;
 
+    /** @var string */
+    public $version;
+
     public $items;
 
     public $currencies;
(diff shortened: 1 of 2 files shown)


app/View/Components/Documents/Script.php
```diff
@@ -45,6 +49,19 @@ public function render()
         return view('components.documents.script');
     }
 
+    protected function getVersion($version)
+    {
+        if (!empty($version)) {
+            return $version;
+        }
+
+        if ($alias = config('type.' . $this->type . '.alias')) {
```
pavel-mironchik, line 58: Pay attention to this part. I don't pass the `$type` variable into the function cause we already have it in the class property. In other blade components, we pass the type mostly everywhere, which I suppose is redundant and just enlarges functions declarations. Would like to know your opinion about this.

PR #1793, opened 2021-01-22
fixed App\Notifications\Sale\Invoice getTagsReplacement() - updated f…

ahmadsyamim:
fixed App\Notifications\Sale\Invoice getTagsReplacement() - updated from invoice_number to document_number


cuneytsenturk:
thanks @ahmadsyamim

PR #1800, opened 2021-01-26
Allow to use custom warning about existing relations in modules.

pavel-mironchik:
With this PR, in a module, we'll be able to provide the text like `employees::general.employees`, and it will be correctly translated. An example: https://github.com/akaunting/module-employees/blob/master/Jobs/Position/DeletePosition.php#L35-L42

Before, it provided something like `general.employees::general.employees` as a translation key, which couldn't be translated, of course.


cuneytsenturk:
Thanks @pavel-mironchik

PR #1811, opened 2021-01-28
onCalculateTotal should be called with timeout

SevanNerse:
onCalculateTotal on document.js should be called with timeout because `items` and `form.items` have to be synchronized.


cuneytsenturk:
Thaks @SevanNerse

PR #1819, opened 2021-01-30
dropzone component is arranged

SevanNerse:


cuneytsenturk:
@SevanNerse Thanks king sevan.

PR #1821, opened 2021-02-01
Format in the datepicker

tognit:
Would it be possible to output the format in the data picker based on the company setting?
As well as the language according to company settings


resources/assets/js/components/AkauntingDate.vue
```diff
@@ -27,6 +27,8 @@
 <script>
 import flatPicker from "vue-flatpickr-component";
 import "flatpickr/dist/flatpickr.css";
+import { de } from 'flatpickr/dist/l10n/de.js';
```
denisdulici, line 30: This shouldn't be loaded by default

resources/assets/js/components/AkauntingDate.vue
```diff
@@ -76,7 +78,8 @@ export default {
 
     data() {
         return {
-            real_model: this.model
+            real_model: this.model,
+            locale: !this.local ? de : en,
```
denisdulici, line 82: What happens if I've set up `fr` in my settings?

denisdulici (changes requested):
Thanks for the contribution, I've added some comments.

denisdulici:
@tognit Any update on this?

tognit:
I'm still there

resources/assets/js/components/AkauntingDate.vue
tognit: The import should be dynamic according to the locale

resources/assets/js/components/AkauntingDate.vue
tognit: in date_group.blade.php you could read out the locale
```javascript
<akaunting-date
.......
      :config="{
            ...........
            locale: '{{ language()->getShortCode() }}',
        }"
..........
></akaunting-date>
```

tognit:
The last update should dynamically adapt the language

PR #1833, opened 2021-02-05
Added `deleted_at` column to recurring check command

burakcakirel:


denisdulici:
This is how it's expected to work, not re-create deleted recurring.

Issue #1856, opened 2021-02-12
Edit transaction is not working on imported transactions


srini85:
When you import transactions, under the transactions area, the edit link routes you to an invalid page

### Steps to reproduce the issue
1. Import transactions from an excel file under the banking, transactions area
2. Click on the date of any of the imported transaction to navigate you to the edit page

### Expected result

Routes you to the edit page

### Actual result

Routes you to 404

![Screenshot_20210213-081127](https://user-images.githubusercontent.com/1079813/107823029-2c147f80-6dd3-11eb-91c2-312aff78d869.png)

### System information (Akaunting, PHP versions)

Akaunting 2.1.1

### Additional comments

Problem is the edit link navigates to is 0 rather than the actual Id of the transaction. In my transaction, it's navigating here

http://accounting.domain.com/sales/revenues/0/edit

If you create a transaction and edit, it works fine. Only seems to be an issue from the imported transaction

srini85:
Looks like issue is due to the import. It Id is null, it should be returning null but it is returning 0 as it casts to int

srini85:
@denisdulici any thoughts on this?

PR #1857, opened 2021-02-13
return null rather than casting to int if the id is null

srini85:
#https://github.com/akaunting/akaunting/issues/1856

Fix to import transaction when we have no reference documents for the transaction. It imports the document_id as 0 rather than null


denisdulici:
What does this exactly fix? No description doesn't help!

srini85:
> What does this exactly fix? No description doesn't help!

Apologies @denisdulici

I had created and issue and referenced the PR there on the issue. Just done it here now and updated the description

srini85:
> > What does this exactly fix? No description doesn't help!
> 
> Apologies @denisdulici
> 
> I had created and issue and referenced the PR there on the issue. Just done it here now and updated the description.
> 
> On a side note-> the tests are failing on the master branch at present. I am kind of new to the whole project and havent got my head around how those tests work. I was just making sure it wasnt something I introduced, but confirmed I had the same issue when running the tests from the master branch prior to my changes.

Just remerged from master as I see you have fixed the tests

PR #1870, opened 2021-02-17
Show a correct text in a delete modal for modules

pavel-mironchik:
This PR adds 2 variables in the DocumentIndex blade component that allow a module to provide options for a delete modal. It also assumes that a translation string is "module_alias::general.document_type", so using this convention we don't need to change the existing modules' code.

The core documents will benefit from this change too - in a delete modal dialog a document number will be shown.

Before:

2021-02-17 09-02-10 Ubuntu Mate 19 10 (Снимок 43)  Работает  - Oracle VM VirtualBox   2

After:

2021-02-17 09-33-48 Ubuntu Mate 19 10 (Снимок 43)  Работает  - Oracle VM VirtualBox   2

### DIFF
--- a/app/Abstracts/View/Components/Document.php
+++ b/app/Abstracts/View/Components/Document.php
@@ -11,7 +11,7 @@ abstract class Document extends Component
     {
         $translation = '';
 
-        // if set config trasnlation config_key
+        // if set config translation config_key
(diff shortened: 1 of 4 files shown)


cuneytsenturk:
Thansk @pavel-mironchik

Issue #1872, opened 2021-02-17
Stylesheets not loading over HTTPS


dev-zm-sl:
Please, submit only real issues. Use the forum for support, feature requests, proposals, new versions, help etc. https://akaunting.com/forum

### Steps to reproduce the issue
Just need a server over https

### Expected result
All styling must work

### Actual result
Stylesheets not loading

### System information (Akaunting, PHP versions)
NginX
PHP 7.4

### Additional comments
![image](https://user-images.githubusercontent.com/78361157/108192025-9744b580-7135-11eb-91ea-955078d005bb.png)

PR #1873, opened 2021-02-17
Update App.php

dev-zm-sl:
Stylesheet issue #1872


denisdulici:
This should be done on server side, not app. You can also add `APP_URL` to your `.env` file.

dev-zm-sl:
its the issue of App not the server as server is loading on HTTPS but the assets are being loaded on HTTP.
I used the APP_URL and ASSET_URL too but of no use

seuaCoder:
It would be nice to fix the ussies before closing them.

PR #1892, opened 2021-02-26
this makes possible adding extra attributes to models

SevanNerse:


pavel-mironchik:
@SevanNerse would you please provide an example of using this feature?

SevanNerse:
@pavel-mironchik For example, there could be an extra field for customers defined with the Custom Fields app. If users want to import customers data with that custom field, it was not possible because that field is not known from the customer model. Hereby, a model created while importing process could move extra field to observer used in the Custom Fields app.

PR #1893, opened 2021-02-27
Fix trying to modify computed value from method

chistel:
In the remoteMethod, when the response does not have data, it tries to modify the compute value "sortOption". this I noticed when I removed the response when testing features on my pc.

also in the remoteMethod, the limit is joined as a search keyword when searching


cuneytsenturk:
Improvements have been made to search remote.
There is no problem in its operation

PR #1905, opened 2021-03-04
Add enlightn to the workflow

paras-malhotra:
This PR adds Enlightn to the workflow as discussed.

For the workflow to work properly (as in https://github.com/paras-malhotra/akaunting/pull/1), we would need to add 2 secrets to the repo: ENLIGHTN_USERNAME and ENLIGHTN_API_TOKEN, and enable the repository on the Enlightn website [here](https://www.laravel-enlightn.com/repositories).


paras-malhotra:
@denisdulici it's still not detecting the secrets I think. Not sure if Github caches secrets at the time of the PR. Should I send across a separate PR and close this one?

denisdulici:
@paras-malhotra That may help as I double-checked the credentials 😞

PR #1915, opened 2021-03-07
Fix Enlightn workflow for secret availability

paras-malhotra:
After some more digging, I found that secrets aren't available to forks even though there is a [proposal under consideration](https://github.community/t/make-secrets-available-to-builds-of-forks/16166).

@denisdulici, I've modified the workflow which should now run in all scenarios (whether or not secrets are accessible). If secrets are available the Pro package will be pulled in and the bot will be triggered. If not available, the Enlightn CI will be triggered without Pro.

It would be interesting to see that if an owner/maintainer triggers a workflow on a fork, whether that makes secrets available.

Nevertheless, this should be a foolproof way for now. Sorry for all the troubles!

### DIFF
--- a/.github/workflows/enlightn.yml
+++ b/.github/workflows/enlightn.yml
@@ -1,9 +1,13 @@
 name: Run Enlightn Checks
 
-on: [pull_request]
+on: [push, pull_request]
 
 jobs:
   enlightn:
+    env:
+      ENLIGHTN_USERNAME: ${{ secrets.ENLIGHTN_USERNAME }}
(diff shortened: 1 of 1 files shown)


denisdulici:
It's working fine now 🎉  https://github.com/akaunting/akaunting/pull/1923

denisdulici:
@paras-malhotra The workflow is failing today, do you have any idea?

paras-malhotra:
@denisdulici, this is due to Composer's security vulnerability: https://blog.packagist.com/composer-command-injection-vulnerability/

Simple fix is: `composer update composer/composer`. Hope this helps!

denisdulici:
That helped, thanks a lot 😉

PR #1936, opened 2021-03-15
Overwrite button styles for ColorPicker element

bengu-thon-mai-mochi:


cuneytsenturk:
Thanks @bengu-thon-mai-mochi

PR #1971, opened 2021-04-05
require laravel-debubar in development only

jonathan-longe:
Suggest moving "laravel-debugbar" into the require-dev section of composer.json as recommended by the package author, "barryvdh".


denisdulici:
Thanks for the suggestion but we may need to put the app into dev mode by just changing the `.env` file.

PR #1978, opened 2021-04-09
If dynamicOptions object has items, then use it.

burakcakirel:


resources/assets/js/components/AkauntingSelectRemote.vue
```diff
@@ -468,7 +468,7 @@ export default {
         },
 
         setSortOptions() {
-            let created_options = (this.dynamicOptions) ? this.dynamicOptions : this.options;
+            let created_options = (this.dynamicOptions.length > 0) ? this.dynamicOptions : this.options;
```
SevanNerse, line 471: the value of `dynamicOptions` might need to be given empty in some circumstances.

PR #1981, opened 2021-04-12
Move document footer to the end of the document

incrolution:
When creating PDF files of bills the footer section is moved to the left besides the invoice summary. This would ensure to keep the footer below the summary

### DIFF
--- a/public/css/print.css
+++ b/public/css/print.css
@@ -488,3 +488,9 @@ th, td
     background-color: #f6f9fc;
     text-decoration: none;
 }
+
+.clearfix:after {
+	display: block;
+	clear: both;
+	content: ""
+}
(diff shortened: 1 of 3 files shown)


cuneytsenturk:
Thank you feedback. Did you check our css file before add custom css class.
Argon theme already this style `clearfix` or `floatfix`
Please check core css then update it.

incrolution:
I added this css class to the print css file since no other css file is used for print views.
Should I rather import the argon css  file to the print export or should i stick with an added css class within the print css file?

cuneytsenturk:
> I added this css class to the print css file since no other css file is used for print views.
> Should I rather import the argon css file to the print export or should i stick with an added css class within the print css file?

This print content use **show** page so then I said can use argon theme class and add `print.css`.

incrolution:
Is this the way you meant it to be?

cuneytsenturk:
Thank you your all effor @incrolution

PR #2011, opened 2021-04-29
Update customer/user relationships

burakcakirel:


cuneytsenturk:
Thanks @burakcakirel tihs good example for request rule.

PR #2048, opened 2021-05-18
Fix the import page title for modules.

pavel-mironchik:
In multi-word routes, we use the `-` as a delimiter, but we use the `_` instead of in the translation keys. So at least for modules, it needs a basic character replacement, which this PR proposes.

Though I see that in the import "xlsx" templates' names the underscore is used, it could be not true for all modules, so this part needs checking before approving this PR, cause the back-compatibility break can happen.

### DIFF
--- a/app/Http/Controllers/Common/Import.php
+++ b/app/Http/Controllers/Common/Import.php
@@ -20,6 +20,7 @@ class Import extends Controller
 
         if (module($group) instanceof \Akaunting\Module\Module) {
             $namespace = $group . '::';
+            $type = str_replace('-', '_', $type);
         } else {
             $namespace = '';
         }


sevannerse:
In my opinion, the names of "xlsx" templates should consist of `-` such as the routes. In this case, change should be done in the [blade](https://github.com/akaunting/akaunting/blob/37d0e7da4e31193a62129abcfb1eb8a838bca352/resources/views/common/import/create.blade.php#L3).

As a result, modules that have template names within `_` should be updated on the next release.

pavel-mironchik:
@sevannerse but then you'll need to change the name of [the import template](https://github.com/akaunting/module-double-entry/blob/master/Resources/assets/chart_of_accounts.xlsx) in your Double Entry module after this PR is merged :)

pavel-mironchik:
By the way, I made a search through the modules' repositories, and it seems that they don't actually use import- my findings show that only Double Entry and Payroll are affected.

sevannerse:
@pavel-mironchik no problem, I will be appreciated to changing the name of the import template file.

The Double Entry was the only module that has faced this situation in the past. That's why I solved it on the module side.

PR #2049, opened 2021-05-18
Fix the import page title for modules (2nd variant).

pavel-mironchik:
Please see the discussion in the #2048


pavel-mironchik:
Could someone check why the Enlightn action didn't pass?

PR #2093, opened 2021-06-08
Named parameter in PHP 8

stoneshaq:
Named parameters were introduced in PHP 8. Parameters without default values must now appear PRIOR to optional parameters. When you use akaunting in PHP 8, it throws an error when you need to view invoices.


cuneytsenturk:
Thanks your PR @stoneshaq

PR #2116, opened 2021-06-18
Allow to use the AkauntingSelectRemote component in a modal dialog.

pavel-mironchik:
This PR allows using of a dropdown with the server-side searching (`AkauntingSelectRemote` component) in a modal dialog. Such an input field is needed in the `Assets` module for [searching/selecting an employee](https://github.com/akaunting/module-assets/blob/master/Resources/views/modals/allotments/create.blade.php#L11). The usual `AkauntingSelect` component is not suitable in this case, as I suppose it's not wise to load all employees into it.


pavel-mironchik:
`composer install` failed while running tests.

PR #2138, opened 2021-06-24
Extract a row of the documents table in a dedicated Blade view.

pavel-mironchik:
This PR not only reduces the number of HTML tags and Blade directives in the `components.documents.index.card-body` view, but also allows modules to inject custom columns in the documents table. If we don't extract the content of a loop, then we don't have the `$item` variable in a ViewComposer, and therefore can't push anything except static strings.

An example of custom column adding in the Expenses module: https://github.com/akaunting/module-expenses/commit/6d75e693cc2728a3262204f164b8bb81f938ea41#diff-446e1e47ff3e465c22a5993fbcfe5933bedf6016ce94fb6434d3402be17c2b67R31-R40


denisdulici:
Task linked: [CU-9g3rhf Employee Name should be displayed on the expense claim listing page](https://app.clickup.com/t/9g3rhf)

PR #2162, opened 2021-06-29
Allow using the dependent validation rules in import.

pavel-mironchik:
This PR provides the ability to use dependent validation rules such as: `after`, `after_or_equal`, `before`, `before_or_equal`, `different`, `exclude_if`, `exclude_unless`, `gt`, `gte`, `in_array`, `lt`, `lte`, `prohibited_if`, `prohibited_unless`, `required_if`, `required_unless`, `required_with`, `required_with_all`, `required_without`, `required_without_all`, `same`.

You just return the import rules like this: `return $this->replaceForBatchRules($rules);`

@cuneytsenturk I've manually checked importing of the sample invoices file and have run a PHPUnit test - both passed fine.

@denisdulici would you please test this too?


cuneytsenturk:
@pavel-mironchik Thank you

PR #2181, opened 2021-07-07
Allow modules to prevent contacts deleting #nka5y8

pavel-mironchik:
This PR allows modules to provide their custom relationships to the `Contact`, thus preventing deleting a contact if there are relationships. Without this, the Core only checks for embedded relationships.

An example from the `Employee` module: https://github.com/akaunting/module-employees/blob/wip/Listeners/PreventContactDeleting.php

### DIFF
--- a/app/Events/Common/ContactGettingRelationships.php
+++ b/app/Events/Common/ContactGettingRelationships.php
@@ -6,10 +6,10 @@ use App\Abstracts\Event;
 
 class ContactGettingRelationships extends Event
 {
-    public $rel;
+    public $contact;
 
-    public function __construct($rel)
+    public function __construct($contact)
     {
-        $this->rel = $rel;
+        $this->contact = $contact;
     }
 }
--- a/app/Events/Common/ContactGettingRelationships.php
+++ b/app/Events/Common/ContactGettingRelationships.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace App\Events\Common;
+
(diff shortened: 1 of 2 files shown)


denisdulici:
Task linked: [CU-nka5y8 Trying to get property 'name' of non-object](https://app.clickup.com/t/nka5y8)

app/Events/Common/ContactGettingRelationships.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace App\Events\Common;
+
+use App\Abstracts\Event;
+
+class ContactGettingRelationships extends Event
+{
+    public $rel;
```
cuneytsenturk, line 9: In order to be more understandable, can you change this variable name to `$contact`?

pavel-mironchik: I agree with you that the variable's name will benefit from renaming, but won't the `$contact` be more confusing for the modules' authors cause it's actually not a contact entity and has no its attributes, etc.?

denisdulici: Agree with both 😃 Therefore, I've sent a more generic solution.

pavel-mironchik: Thanks, @denisdulici 
Honestly, I hoped that my PR will inspire the Core's devs to implement a unified solution, as you provided.

PR #2190, opened 2021-07-14
fix: Update module edit form to allow number input

phpsa:
This adds the ability to use a number input on module settings page.


cuneytsenturk:
Thanks, you effor

PR #2200, opened 2021-07-27
Fetch items only after default character limit is reached

bengu-thon-mai-mochi:
Dropdown for list of items does ajax call to search for items only after default character limit is reached.

I have provided the default to 3 at the moment and will integrate with back-end data once the data is available. 

For example, if the user is searching for 'notebook', it will only do ajax call after 'not' is written by the user.  

Screen Shot 2021-07-27 at 09 28 08


cuneytsenturk:
Thank you for your all effor @bengu-thon-mai-mochi

PR #2223, opened 2021-08-11
Adding discount as an amount

burakcakirel:


cuneytsenturk:
@burakcakirel Thanks for all your effort.

PR #2249, opened 2021-09-11
Get source from the request if the document's item has not item_id

burakcakirel:


denisdulici:
Can you plz delete lines 200-201?

PR #2253, opened 2021-09-16
Create SECURITY.md for security researchers.

zidingz:


denisdulici:
Thanks but not interested in it. Please, don't insist.

amammad:
Hello my dear denisdulici
I don't want to insist, as a security researcher I just want to say that the SECURITY.md or same Security policy is a standard procedure that can give every good repository like akaunting more good reputation and Also there is no need to always change it you can just say 'we accept security vulnerabilities for last stable version and this is our confidential email for contacting'

Best regards.

PR #2277, opened 2021-10-06
Show category name instead of id in remote select

bengu-thon-mai-mochi:


cuneytsenturk:
This issue already fixed so then ı am closing it.

PR #2313, opened 2021-11-17
Missing country, city, etc columns for import files.

CihanSenturk:


denisdulici:
Task linked: [CU-1t1uf4t Contact_Country column is missing in sample import files](https://app.clickup.com/t/1t1uf4t)

PR #2337, opened 2021-12-11
Fix timeout

shandak:
When your table `document_items` has around 65k records and you try to filter invoices in view '.../sales/invoices' then the server returns a timeout error, because it selects all unnecessary dependencies for the current request.

### DIFF
--- a/app/Http/Controllers/Sales/Customers.php
+++ b/app/Http/Controllers/Sales/Customers.php
@@ -26,7 +26,7 @@ class Customers extends Controller
      */
     public function index()
     {
-        $customers = Contact::with('invoices.transactions')->customer()->collect();
+        $customers = Contact::query()->customer()->collect();
 
         return $this->response('sales.customers.index', compact('customers'));
     }


denisdulici:
Thanks for your contribution, but that's required to calculate the `Unpaid` value. It's paginated, so it won't load all records.

shandak:
Well, I did not say that it was not paginated. I have just around 10 customers which have related around 65k records in `document_items` so there is nothing to paginate. Why it should calculate Unpaid if there is only a selected list of customers for filter purposes in view '.../sales/invoices'? This query is enormously expensive and I suppose there can be found a different way to calculate `Unpaid` if you need it.

denisdulici:
Well, you didn't say that those 65K records were related to 10 customers, so yes, the records are paginated. Feel free to send a PR for a different way to calculate the `Unpaid` value.

shandak:
I don't really get when you say "calculate the Unpaid value". Give me please more details about where it's used and which conditions must be in the query.

PR #2338, opened 2021-12-14
fix console error in not found locale

ali00h:
If we select a locale which does not exist in flatpickr locales, an error is displayed in the user browser.
I added a try,catch block to fix this error.


cuneytsenturk:
Thanks for your feedbacks.

PR #2339, opened 2021-12-14
json_encode thousands seperator so js won't break

josias-r:
Before:
generates invalid JS like this:
```
const moneySettings = {
   decimal: '.',
   thousands: ''', // THIS IS INVALID
   symbol: 'CHF',
   isPrefix: '1',
   precision: '2',
};
```
Causes the "Umlaufvermögen" Chart to break in the FrontEnd on the Dashboard. (Shows infinite loading screen)

You probably also want to json_encode all the other possible settings, (or some other processing of the variable) that are patching together the JS. The `thousands` is just really susceptible to this bug because `'` is a common thousands seperator.

After:

Screenshot 2021-12-14 at 16 14 36

### DIFF
--- a/app/Traits/Charts.php
+++ b/app/Traits/Charts.php
@@ -145,7 +145,7 @@ trait Charts
                     'label' => new Raw("function(tooltipItem, data) {
                         const moneySettings = {
                             decimal: '" . config('money.' . setting('default.currency') . '.decimal_mark') . "',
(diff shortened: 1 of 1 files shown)


josias-r:
@sevannerse Will this PR get noticed, or do I have to open an issue for it? :D

cuneytsenturk:
I've solved this commit
https://github.com/akaunting/akaunting/commit/da5f1988c2d01bee4bb9138a3c996ee197a1ce44

PR #2353, opened 2021-12-30
Time out fix

shandak:
When your table document_items has around 65k records and you try to filter invoices in view '.../sales/invoices' then the server returns a timeout error, because it selects all unnecessary dependencies for the current request.
The previous solution  https://github.com/akaunting/akaunting/pull/2337 was declined

### DIFF
--- a/app/Http/Controllers/Sales/Customers.php
+++ b/app/Http/Controllers/Sales/Customers.php
@@ -26,7 +26,7 @@ class Customers extends Controller
      */
     public function index()
     {
-        $customers = Contact::with('invoices.transactions')->customer()->collect();
+        $customers = Contact::query()->customer()->collect();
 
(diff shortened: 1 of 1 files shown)


pavel-mironchik:
Doesn't the `Contact::has('invoices.transactions')` retrieve only those customers that have paid their invoices fully or partially? So you won't see in the Customers list the customers that don't have invoices or didn't pay their invoices yet. Or did I miss something?

shandak:
It looks like, yes. As I could understand there https://github.com/akaunting/akaunting/pull/2337 is a requirement. IMO it's not really useful because we already have a filter by status. And, well, at least my fix does not change logic but improves.

denisdulici:
Closing for the same reason.

PR #2360, opened 2022-01-21
Allow using custom labels for a radio button

pavel-mironchik:
For now, the "radio" component shows only the default "yes/no" variants regardless of passed attributes:

2022-01-21 15-39-11 Edit Currency - Modules Testing - Google Chrome

https://github.com/akaunting/akaunting/blob/master/resources/views/partials/form/radio_group.blade.php#L18
https://github.com/akaunting/akaunting/blob/master/resources/views/partials/form/radio_group.blade.php#L30

This PR changes the component's behavior so it will use the provided strings for labels, but will back to defaults if there are no custom labels. Thus, it won't break any existing code in the Core and modules.


cuneytsenturk:
Thanks @pavel-mironchik

PR #2364, opened 2022-01-25
Add Support icon to Mobile version

leonardofilhogm:


denisdulici:
Fixed in Akaunting 3.0

PR #2373, opened 2022-03-09
Fix to show persisted select option in app settings

dadeke:
Fix to show persisted select option in app settings.

Example:

PayFlexi - [https://akaunting.com/apps/payflexi](https://akaunting.com/apps/payflexi)

screenshot_settings


dadeke:
Closed for reorganization. 🙂

PR #2380, opened 2022-03-28
Extra documents created on recurring

neslonso:
[-] The two calls to recur with different dates cause skipThisClone to be called each time, which will return false in one of them (unless there are both clones, one for each date, which doesn't make much sense). This creates extra clones. Now all the schedules are traversed and recur is called for those that do not already have a clone on one of the two dates.


denisdulici:
Fixed in Akaunting 3.0

PR #2396, opened 2022-05-10
Fix issue in translation `ar-SA`

michaelnabil230:
- Fix the translation and `HTML` tag

### DIFF
--- a/resources/lang/ar-SA/settings.php
+++ b/resources/lang/ar-SA/settings.php
@@ -111,7 +111,7 @@ return [
         'templates' => [
             'subject'                   => 'الموضوع',
             'body'                      => 'الجسم',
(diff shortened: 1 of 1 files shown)


denisdulici:
You can use Crowdin for translations https://crowdin.com/project/akaunting

michaelnabil230:
Hi @denisdulici 
This is a problem in the translation, it is not a simple thing missing in the translation, it will disable something else in the html tags when the render in browser and also there will be a problem with the directions when you add the rtl style

michaelnabil230:
Also, if you open the code, you will see that there is a wrongly closed html tag

michaelnabil230:
@denisdulici 
Please pay attention

PR #2397, opened 2022-05-12
Update to RTL support, based on PR 2212

arnonm:
This improves on the RTL support PR. It includes the changes to _rtl.scss and adds helpers to determine rtl languages.
These helper functions should probably be moved to akaunting/laravel-language but that would require another PR


denisdulici:
Available in Akaunting 3.0

PR #2410, opened 2022-06-02
Updated sl-SI language to match folder

Markoftw:
Updated the sl-SI language in the config to match the folder.
Closed <strong> HTML tag.


denisdulici:
Thanks @Markoftw

PR #2424, opened 2022-06-06
Fix 500 error in Sales/Customers view mode

burakcakirel:
#30xc7aw


denisdulici:
Task linked: [CU-30xc7aw App is causing 500 error in Sales/Customers view mode](https://app.clickup.com/t/30xc7aw)

PR #2482, opened 2022-06-25
Replace function `getTimezones` in to collect

michaelnabil230:


EnesSacid-Buker:
Thanks @michaelnabil230 for contributing

PR #2527, opened 2022-07-13
Added missing @endif

StefanArts:
Adding the missing @endif stops 500 Error when visiting App Info


cuneytsenturk:
This PR broken control so then ı can not merge it. Thanks your effort

PR #2540, opened 2022-07-19
Create new if import has no currency code

CihanSenturk:


denisdulici:
Task linked: [CU-37h9uwk Import currency_code fields must be firstOrCreate](https://app.clickup.com/t/37h9uwk)

PR #2578, opened 2022-08-04
Categories import sample file type field development

CihanSenturk:
image


denisdulici:
Task linked: [CU-39nxnyc Type dropdown in Category example excel file](https://app.clickup.com/t/39nxnyc)

PR #2585, opened 2022-08-08
Uses database config from env while install

yasirmturk:
Currently it is impossible to load database config values from env file


EnesSacid-Buker:
Thanks @yasirmturk

PR #2586, opened 2022-08-08
Uses locale config from env while install

yasirmturk:
Currently it is impossible to load database config values from env file
Also, if user is not coming from `install/language` then locale is missing in session


EnesSacid-Buker:
Thanks @yasirmturk

PR #2588, opened 2022-08-09
Fix: Download file should be possible regardless delete permissions.

desplega:
Currently the "delete-common-uploads" permissions are blocking the possibility to download an attached file for a user who does not have "delete" permissions. It is proposed to limit the effect of this limitation to the "delete" action, allowing to donwload the file.


cuneytsenturk:
Thank for PR

PR #2595, opened 2022-08-11
Fix: Use proper translations from module in Portal

desplega:
Module translations in Portal page for the Empty page (when no records in the module have been created yet) are not correct. The problem is that inside the EmptyPage.php file the module is detected as "portal.modulename" and translations are not found. I propose to remove the "portal." section of the $alias so that the module is correctly detected and the proper translations are used.
I detected this problem through Helpdesk module, which is used in Portal, however it should be tested with other modules to ensure the proposed fix works for all of them.

### DIFF
--- a/app/View/Components/EmptyPage.php
+++ b/app/View/Components/EmptyPage.php
@@ -82,6 +82,9 @@ class EmptyPage extends Component
             $alias = $group;
         }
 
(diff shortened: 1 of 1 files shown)


cuneytsenturk:
@desplega Marcel you can use `alias` not necessary these changes. Thank your effort.

PR #2631, opened 2022-09-19
Add Dockerfile to simplify installation

0x1a8510f2:
This PR adds a `Dockerfile` to the root of the repo in order to simplify installation (allow installing as a container) for people who may not have a LAMP stack or equivalent installed directly on their machine.

There are probably some tweaks which could be made to improve the resulting container but I wanted to get an idea first if this is something you'd be interested in.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM debian:bookworm
+
+WORKDIR /var/www/html
+
+RUN apt-get update && \
+        apt-get upgrade -y && \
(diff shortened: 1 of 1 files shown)


denisdulici:
Thanks for your contribution but there is a separate repo for that purpose https://github.com/akaunting/docker

0x1a8510f2:
Ah, completely missed that. It may be worth mentioning that repo in the docs though because, at the moment, it's extremely difficult to find:

image

Issue #2702, opened 2022-10-27
Install wizard not working in Firefox


Androggles:
### Akaunting version

3.0.7

### PHP version

8.1.11

### Operating system

Debian 11 (Turnkey LAMP)

### Steps to reproduce

Upload files to webserver, and start the install wizard.

When clicking next on the language selection, it will just redirect to the same page.

### Expected result

Should be able to go throught the install wizard.

### Actual result

Redirects to step one.

### Additional comments

Works if doing it in Google Chrome or Edge.
Nothing shows up in the logs in either Apache or Akaunting.

PR #2703, opened 2022-10-28
Added prevent.default() for Install form

brkcvn:
Firefox form posting error fixed. Because` prevent.default()` is not, url couldn’t redirect when form posting on Firefox..


denisdulici:
Fixed #2702

PR #2738, opened 2022-11-10
Multi select value improvement 

brkcvn:
When you use form post into the select component , previous value was disappearing. Fixed this issue.


brkcvn:
This development doesn’t works fine so I will delete this task

PR #2873, opened 2023-01-17
use the correct translation for cancellations

novag:
Some languages have different words for cancelling an action vs cancelling an invoice. Currently, the label for cancelling an invoice is very confusing in those languages.

This pull request introduces a separate string for the action 'cancel invoice/bill'.

The 'mark_cancelled' strings in 'lang/.../bills.php' and 'lang/.../invoices.php' may have been introduced earlier for this purpose. In the meantime, however, the buttons use the generic "cancel" string, which is why the problem occurs.

Ref.: #2868 was closed by mistake.


novag:
Rebased

PR #2876, opened 2023-01-18
align document cancellation process with create/update/delete flow

novag:
Allows modules to hook into different stages of the cancellation process and present error messages to users.


EnesSacid-Buker:
hi @novag

Your improvement is nice, but in modules that use the document structure, we need to check the same way and put the other listeners in the same format that change state in the other document.
We took note of this PR.
Thank you for contributing

PR #2878, opened 2023-01-23
Direct server.php to index.php at the app root

nnq1:
index.php of **Akaunting is in the root** of it. but
server.php which is in the: vendor/laravel/framework/src/Illuminate/Foundation/resources/server.php
directs the app to index.php in the public directory by:
require_once $publicPath.'/index.php';
but I modify the line to:
require_once $publicPath.'/../index.php';
Thus, it will use index.php in the App root.
Of course, we can move index.php to the public folder which is a convention of Laravel project, but with Akaunting it produce other errors!


EnesSacid-Buker:
Hi @nnq1
 
You can use this instead of artisan serve
```
# if console location is current working directory
php -S localhost:8000
# if is not
php -S localhost:8000 -t {working directory}
```

PR #2885, opened 2023-02-12
check media existence considering the disk

novag:
Fixes downloads for media entries located on a disk other than the one specified in the FILESYSTEM_DISK environment variable.

This allows modules to manage media data on their own disks.

The 'allowed_disks' list specified in the mediable configuration still applies. (https://github.com/akaunting/akaunting/blob/master/config/mediable.php#L26)

### DIFF
--- a/app/Traits/Uploads.php
+++ b/app/Traits/Uploads.php
@@ -112,7 +112,7 @@ trait Uploads
 
         $path = $media->getDiskPath();
 
-        if (Storage::missing($path)) {
+        if (Storage::disk($media->disk)->missing($path)) {
             return false;
         }
 


EnesSacid-Buker:
hi @novag 

I tested with [MinIO](https://min.io) your code but got the same result without adding the disk function.
Maybe something is missing in your .env
my .env 👇🏼

EnesSacid-Buker:
hi @novag 

I tested your code with [MinIO](https://min.io) & removed s3 from allow_disks but got same result without adding disk function.

novag:
Hello @EnesSacid-Buker,

the problem is not about the FILESYSTEM_CLOUD (s3) disk but for additional disks added and managed by modules/apps.
You can test this by dynamically adding a new disk to the filesystems settings from a module. Then upload a file to this disk and try downloading it.

EnesSacid-Buker:
Hi @novag 

I understood what you mean.
I fixed this problem in other parts of the software.
Thank you for noticing such an error and reporting it to us.

novag:
Hi @EnesSacid-Buker,

thanks for getting back! After reviewing your changes, I agree that they are indeed more appropriate.

PR #2891, opened 2023-02-16
The issue of getting media from different disks

EnesSacid-Buker:
### This bug was noticed with this #2885


denisdulici:
Task linked: [CU-8669j43b2 Use the media model instead of storage facade](https://app.clickup.com/t/8669j43b2)

PR #2904, opened 2023-02-24
introduce 'unit' line item column

novag:
In many industry sectors it is required to specify units for line items. This pull request adds a unit column and allows the user to customize its name or hide the column.


novag:
Rebased.

novag:
Rebased.

novag:
Rebased and fixed a bug that prevented the unit column from being hidden if configured.

cuneytsenturk:
Thank for feed back @novag  but,
Already this column will be added [Inventory](https://akaunting.com/apps/inventory) app, So then ı can not merge this feature on the core

novag:
@cuneytsenturk Please reconsider the introduction to Akaunting core. This column is necessary not only for goods management, but also for services to distinguish, for example, between hourly and daily rates or other modalities.

PR #2909, opened 2023-02-26
allow hooking functions related to document numbers

novag:
This change allows modules to hook the document number generation functions through the Laravel 'Macroable' interface.

All changes are backwards compatible and do not change current behavior.


denisdulici:
Unfortunately, can't accept this as it's not hooking but overriding the functions.

PR #2911, opened 2023-02-28
Do not show Paid when is Unpaid

dadeke:
Showing Paid when is Unpaid


cuneytsenturk:
Thanks for feedback we can use diffrent method.

PR #2953, opened 2023-03-27
Fix special characters for tax in templates

markoftw:
This fixes the **green text** and **special characters** for Tax fields in templates, previews & PDF downloads.

Before:

Web preview
image

PDF
image

After:

Web preview
image

PDF
image


cuneytsenturk:
Thanks @markoftw.

PR #2970, opened 2023-05-21
add 'inline' route to allow embedding uploads

novag:
This change adds an 'inline' route to uploads which sets the
'Content-Disposition' HTTP header to 'inline'. This allows modules to
embed and display uploads in an HTML 'embed' element.


cuneytsenturk:
Thanks for pull request @novag

PR #2984, opened 2023-06-13
Add category object and category_id to document resource

vitormattos:
The endpoint api/documents don't return the category of invoice/bill as the endpoint api/transactions do. With this change will be possible return this value.

The return after this change:

```json
{
	"data": [
		{
			"id": 3,
			...
			"category_id": 6,
			"category": {
				"id": 6,
				"company_id": 1,
				"name": "Category name",
				"type": "expense",
				"color": "green-500",
				"enabled": true,
				"parent_id": null,
				"created_from": "core::ui",
				"created_by": 1,
				"created_at": "2023-06-13T04:21:41+01:00",
				"updated_at": "2023-06-13T04:21:41+01:00"
			},
```


cuneytsenturk:
thanks for this feedback @vitormattos

PR #3000, opened 2023-07-04
Left menu chevron not rotating

mervekaraman:


cuneytsenturk:
Thanks @mervekaraman and this PR number #3000 🚀 Akaunting is getting stronger every day 💪

PR #3075, opened 2023-10-18
Fixed dashboard bulk action issues

CihanSenturk:


denisdulici:
Task linked: [CU-8678xq6nh The enable and delete features in the bulk action menu do not work. gives an error. applies to dashboards that are disabled](https://app.clickup.com/t/8678xq6nh)

PR #3093, opened 2023-11-12
Fixed missing default in settings match

jonas32:
Fixed bug where finding the env key for settings fails after last update's change from switch to match causing settings to be unchangeable


cuneytsenturk:
Thanks  @jonas32

Razva:
Is this merged into `3.1.4` or have to wait 'till `3.1.5`?

PR #3229, opened 2024-10-14
Added footer css to set footer in the bottom of the page

Js-Developer-web:
The footer should always be at the bottom of the page. So, Here I added some styling to the footer


cuneytsenturk:
I can't accept this PR because the app.css under public. is dynamically updated. The change you make here will go in the next version. For this, make sure you make this change in the correct location.

PR #3236, opened 2024-11-06
Add PHP 8.4 to workflows.

Maikuolan:


Maikuolan:
(Fixed and resynced with master).

Maikuolan:
Okay.. There's a bit of problem with accepting this, unfortunately. According to the failed test:

```
Run composer test
> composer install --prefer-dist --no-interaction --no-scripts --no-progress --no-ansi
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - brianium/paratest is locked to version v7.3.1 and an update of this package was not requested.
    - brianium/paratest v7.3.1 requires php ~[8](https://github.com/akaunting/akaunting/actions/runs/13734547294/job/38416601097?pr=3236#step:8:9).1.0 || ~8.2.0 || ~8.3.0 -> your php version (8.4.4) does not satisfy that requirement.

Error: Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - brianium/paratest is locked to version v7.3.1 and an update of this package was not requested.
    - brianium/paratest v7.3.1 requires php ~8.1.0 || ~8.2.0 || ~8.3.0 -> your php version (8.4.4) does not satisfy that requirement.

Script composer install --prefer-dist --no-interaction --no-scripts --no-progress --no-ansi handling the test event returned with error code 2
Error: Process completed with exit code 2.
```

So, it seems, the current dependencies constraints require a PHP version lower than PHP 8.4, effectively meaning they're incompatible with PHP 8.4, and attempting to update those constraints in order for them to be compatible with PHP 8.4 would likely render them incompatible with previous PHP versions. Not sure the best way to approach that problem.

Considering the problem, maintainers possibly may want to reject this pull request. As a non-maintainer and the pull request's author, I'll leave it open for the possibility of it being merged, as the decision is ultimately up to the maintainers. But, I won't be offended at all if it's rejected, particularly as there's a good reason to reject it, considering the problem at hand. :-)

Perhaps, we just regard v3 and earlier of akaunting to be incompatible with PHP 8.4 (as removing or swapping out any dependencies could be considered as a BC-breaking change, and if considered that way, then per semver guidelines, would warrant the next release from then to be at least v4), and then earmark PHP 8.4 compatibility as being one of the goal's for a future v4 release? Might be able to retain backwards compatibility with older PHP versions while also bringing in PHP 8.4 at the same time if there exists a comparable PHP package or available branch of the relevant affectant dependencies which said affectant dependencies could be swapped out with, but I haven't investigated that yet. If not, then likely, bringing in PHP 8.4 compatibility would likely break compatibility with previous PHP versions, in which case, it probably shouldn't happen for anything prior to v4.

Neustradamus:
@Maikuolan: Why have you removed your PR?

Maikuolan:
> @Maikuolan: Why have you removed your PR?

The pull request is only closed, not removed. The change (which was only 1 line anyway and easy to replicate) can still be reviewed via the "files changed" tab, and if need be, the relevant branch restored and eventually merged at the maintainer's discretion if they so choose. (I would interpret a "removal" in this context as something like asking GitHub to delete the pull request or making it entirely inaccessible, which would be a far more extreme action to initiate than simply closing it and not something necessary here).

The reason for closing it is two part: First, that the problem I mentioned earlier is still in effect (`brianium/paratest` dependency constraints have been updated since the time I'd closed this pull request, but the new constraints still require a version, 7.3.2, which itself has PHP version constraints requiring a PHP version >= 8.1 <= 8.3, which means adding PHP 8.4 to the workflow in question here would, as the project's current composer dependencies are at this current time, still result in continuous failures in the event of being approved/merged; there are some much newer versions of `brianium/paratest` which seem to support PHP 8.4, but which also remove support for some older PHP versions which `akaunting/akaunting` still supports in its current major version, so I foresee it potentially being a bigger headache for devs/maintainers than worrying about a one-line pull request from an external contributor is likely worth, in the sense that it would raise questions about whether to drop support for older PHP versions in order to support newer versions as per the PHP version constraints of the dependency in question, whether that would force them to suddenly need to tag a new major version outside of any existing plans there might be for that due to semantic versioning rules, whether to drop support for that dependency entirely, what changes would need to be made in order to accommodate potentially dropping that dependency, whether that might break things elsewhere and so on). I *suspect* that it likely also why none of the maintainers have actually replied to this specific pull request despite its age.

Second, putting aside this specific pull request, the project as a whole is still being actively maintained, issues and pull requests and such still being addressed, commits still regularly pushed, etc. The longer the pull request is left open, the greater the distance between the head of the pull request and the head of the master branch here grows, which IMO is kind of messy.

All that in mind, my thinking was (and still is) that it's better to just close the pull request, for me to leave the issue alone (at least for now), and see what the maintainers do. If `akaunting/akaunting` reaches a state where testing against PHP 8.4 won't cause workflow failures (in whichever shape that state may take, which, as a non-maintainer, is not up to me to decide upon), assuming they don't just update the workflow themselves at that point (it *is* only a one-line change, after all; once it's ready to happen, it's easy enough for anyone to commit that), it would be a simple enough matter for me to just create a *new* pull request with the relevant change, forking from a cleaner, more up-to-date head than this pull request.

In short: Creating this pull request in the first place was a bit premature on my part, and if I had've caught the aforementioned dependency problem beforehand, I would've not created it to begin with.

PR #3252, opened 2025-01-10
add MariaDB to README.md

robertsilen:
Any reason not to include MariaDB in the README? :) Suggesting to add.


robertsilen:
I mean, for example in this issue from a year ago, MariaDB is even working better than MySQL: https://akaunting.com/forum/discussion/installation-update/database-connection-issue-during-install

It would be fair to at least mention MariaDB in the readme, or even recommend it? :) Thanks

Issue #3315, opened 2025-11-05
Recurring Invoices Break When Creating or Updating


imhayatunnabi:
### Akaunting version

3.1.19

### PHP version

8.2

### Operating system

macOS Version 15.6 (24G84)

### Steps to reproduce

Create or edit a recurring invoice from the admin UI (or hit the corresponding `/recurring-invoices` API endpoint).
Submit the form with valid data.

### Expected result

Recurring invoice saves successfully and retains all supplied fields (contact, items, schedule, etc.)

### Actual result

Job receives an empty payload, so the document is stored without the submitted attributes, yielding validation/database errors or a corrupted recurring invoice record.

### Additional comments

Merge the overridden dates first, then pass the actual request instance into the job:
  ```php
  $request->merge(['issued_at' => $request->get('recurring_started_at')]);

  $response = $this->ajaxDispatch(new CreateDocument($request));
  ```
  Apply the same pattern in `update()`.

### Relevant log output

```shell

```

PR #3316, opened 2025-11-05
Recurring Invoices Break When Creating or Updating

imhayatunnabi:
Issue Resolved: 
- When creating or editing a recurring invoice, `RecurringInvoices::store()` and `RecurringInvoices::update()` dispatch `CreateDocument` / `UpdateDocument` jobs by calling `new CreateDocument($request->merge([...]))`.
- `FormRequest::merge()` mutates the request but returns `void`, so the expression ends up passing `null` into the job constructors.
- The job boot logic sees no request payload, builds an empty mock request, and the invoice gets created/updated without any of the validated fields (type, contact, items, schedule), leading to errors or corrupt data across the recurring invoices feature and any integrations that rely on it.

https://github.com/akaunting/akaunting/issues/3315

### DIFF
--- a/app/Http/Controllers/Sales/RecurringInvoices.php
+++ b/app/Http/Controllers/Sales/RecurringInvoices.php
@@ -80,7 +80,9 @@ class RecurringInvoices extends Controller
      */
     public function store(Request $request)
     {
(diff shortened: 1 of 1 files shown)


cuneytsenturk:
Thanks for your help

PR #3317, opened 2025-11-05
Fix missing non-recurring helpers in shared traits

imhayatunnabi:
Enhancement Done:
- replace the `isNotRecurringDocument()` and `isNotRecurringTransaction()` guards to call their existing recurring helpers instead of the undefined `isRecurring()`.
- ensure every document/banking workflow can safely ask whether a record is non-recurring without triggering a fatal `BadMethodCallException`.


cuneytsenturk:
Thank for help

PR #3318, opened 2025-11-05
Enhance Bangla locale coverage and terminology for missings keys and files

imhayatunnabi:
As a Bangladeshi Bengali native speaker and active developer using Akaunting in production, I noticed several gaps in the Bangla experience. This pull request realigns the bn-BD locale with the current en-US strings, adds the missing transations

#### Bangla language enhancement summary

 - Synced every bn-BD locale file with the current en-US structure, filling in the new keys and refreshing
    translations where meaning shifted
  - Added Bangla coverage for the locales that were previously missing by introducing fresh files matching the file with en-US.
  - Rewrote the validation messages to include Laravel’s latest rules while keeping the tone consistent


cuneytsenturk:
You can use Crowdin for translations https://crowdin.com/project/akaunting

PR #3356, opened 2026-04-09
Performance optimization reduce memory usage in income totals database-level aggregation

krornig-kaos:


cuneytsenturk:
@krornig-kaos Thank you for your feedback.

PR #3368, opened 2026-06-29
Fix bill payment email template

Glucksberg:
## Summary

- Use the vendor payment template when sending emails for bill payment transactions.
- Add the missing en-US banking payment email template strings used by seeded transaction templates.
- Cover the bill document payment template mapping with a focused feature test.

## Root cause

`Document::BILL_TYPE` was configured with `invoice_payment_customer`, so using Save & Send after recording a bill payment could open a customer-facing invoice receipt email that says a payment was received. Bills represent vendor payments, and expense transactions already map to `payment_made_vendor`.

## Validation

- `php -l config/type.php`
- `php -l resources/lang/en-US/email_templates.php`
- `php -l tests/Feature/Purchases/BillsTest.php`
- `git diff --check`

I attempted `php artisan test --filter=ItShouldUseVendorPaymentTemplateForBillPayments`, but this checkout initially had no `vendor/autoload.php`. A local Composer install was then blocked by missing PHP extensions in the environment: `bcmath`, `dom/xml/simplexml`, `gd`, `intl`, and `zip`.

### DIFF
--- a/config/type.php
+++ b/config/type.php
@@ -257,7 +257,7 @@ return [
             'contact_type'              => Contact::VENDOR_TYPE,
             'inventory_stock_action'    => 'increase', // increases stock in stock tracking
             'transaction' => [
-                'email_template'        => 'invoice_payment_customer', // use email template
+                'email_template'        => 'payment_made_vendor', // use email template
             ],
             'hide'                      => [],
             'notification' => [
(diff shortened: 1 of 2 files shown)


cuneytsenturk:
Thanks @Glucksberg

Issue #1618, opened 2020-09-24
Item with Invalid or Comma Separated Sale or Purchase Price Saves Incorrectly


SethFalco:
### Steps to reproduce the issue
1. Open the Items page.
2. Click "+ Add New" to create a new item.
3. 
    * Complete the form, but for "Sale Price", or "Purchase Price", put in any character that isn't `[0-9,.]`, for example `£36`.
    * Complete the form, but for "Sale Price", or "Purchase Price", put a number with the thousand separator, such as `1,000` for £1,000. 
4. Save the item and observe the new row in the items table.

### Expected result
The form could either:
* Accept the input if it's logical, for example £36 instead of 36, or at least mark this as invalid.
* Display an error to the user that the form was not filled in correctly.

### Actual result
Where an invalid character is used, the item is saved successfully, but the Sale Price and Purchase Price becomes 0 instead of the value specified by the user.

Where a thousand separator is used, it only reads up to the first comma, for example: `1231.13,21313,13213.12313,1313` is read as `1,231.13` and the rest seems to be ignored, and this value is saved to the item _successfully_.

### System information (Akaunting, PHP versions)
Cloud Instance at app.akaunting.com
Version: 2.0.23

Samkit-shah:
@denisdulici can i take this and enable inputs just as number? @denisdulici

PR #3372, opened 2026-07-13
Fixed thousand-separator parsing for item sale/purchase price

TowyTowy:
The `Money` middleware normalizes localized `sale_price`/`purchase_price` input by replacing commas with dots and then collapsing the thousand separators, keeping only the last dot as the decimal mark. An operator-precedence mistake assigned the boolean result of the comparison to `$dot_count` instead of the separator count:

```php
if ($dot_count = Str::substrCount($money_format, '.') > 1)
```

Since `>` binds tighter than `=`, `$dot_count` was always `true`/`false`, and the inner `$dot_count > 2` branch — which handles values with three or more separators — was dead code. As a result prices such as `1.234.567,89` were truncated to `1234.567` when saved, the exact thousand-separator issue this code was added to fix (#1618). Wrapping the assignment in parentheses restores the intended count.

Verified before/after against the real middleware class: `1.234.567,89` → `1234567.89` and `12.345.678,90` → `12345678.9` now parse correctly, while the existing two-separator (`1.234,56`, `1,234.56`) and plain-number cases are unchanged.

_AI-assisted (Claude) and reviewed/verified before submitting._

### DIFF
--- a/app/Http/Middleware/Money.php
+++ b/app/Http/Middleware/Money.php
@@ -50,7 +50,7 @@ class Money
             if ($parameter == 'sale_price' || $parameter == 'purchase_price') {
                 $money_format = Str::replace(',', '.', $money_format);
 
-                if ($dot_count = Str::substrCount($money_format, '.') > 1) {
+                if (($dot_count = Str::substrCount($money_format, '.')) > 1) {
                     if ($dot_count > 2) {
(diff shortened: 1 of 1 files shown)


cuneytsenturk:
@TowyTowy Thank you for your feedback.

Issue #3319, opened 2025-11-09
Asset loading error through php artisan serve, version requirements needed and docker setup


kaiji123:
### Akaunting version

3.1.19

### PHP version

8.2

### Operating system

Microsoft Windows 10 Pro - Version	10.0.19045 Build 19045

### Steps to reproduce

**Asset Loading**
When using php artisan serve for local development, CSS and JavaScript assets fail to load, mainly because in the code they are accessing through `/public/css/... `instead of `/css/.....`
Therefore the whole styling broke. I suggest mentioning this in `README.md`. 

A quick workaround was to create a symbolic link at public folder linking to itself

for example in powershell 
```
cd public
cmd /c mklink /D public .
cd ..
```

**Versioning** 
It would be also great to mention version requirements for php, node in `README.md`. When using node v22.13.1 (64-bit), I will usually get the gyp error during npm install , so had to change to version v18.18.0 (64-bit) to resolve this issue. The same version compatibility problem happened for php version, was using 8.4 but received an error, so had to change to 8.2.

**Docker Setup**
I have prepared a simple docker compose setup for mysql 

```yml
version: '3.8'

services:
  mysql:
    image: mysql:8.0
    container_name: akaunting-mysql
    restart: always
    environment:
      MYSQL_ROOT_PASSWORD: pass
      MYSQL_DATABASE: akaunting
    ports:
      - "3307:3306"
    volumes:
      - mysql_data:/var/lib/mysql

volumes:
  mysql_data:

```

but for this to work, you HAVE TO increase the time limit in `Installer.php` to use php artisan install script, otherwise you would get a time exceeded error, and then rerunning installation again, would make it fatal error, because of some table constraint (since it got interrupted in the middle). 

```php
  // Try to increase the maximum execution time
        set_time_limit(8000); // 5 minutes
```

### Expected result

<img width="1228" height="686" alt="Image" src="https://github.com/user-attachments/assets/66a63405-b194-4527-a2c5-43cb5f799c63" />

### Actual result

<img width="1423" height="775" alt="Image" src="https://github.com/user-attachments/assets/b52f1e27-fc34-40fc-a7f2-fa0529d986d7" />

<img width="1919" height="333" alt="Image" src="https://github.com/user-attachments/assets/5d7b32ff-c925-4c07-a7a3-07c23e182ae2" />

[...]

PR #3378, opened 2026-07-25
Fixed database host and port not being applied to the current connection on install

Amadeus-22:
### Problem

Installing against a database on a non-default host or port fails during the migration, even though the installer has just accepted the same credentials:

```
Setting locale
Creating database tables
Connecting to database akaunting@127.0.0.1:3307

   Illuminate\Database\QueryException

  SQLSTATE[HY000] [1045] Access denied for user 'root'@'localhost' (using password: YES)
  (Connection: mysql, SQL: select table_name as `name` ... from information_schema.tables ...)

  at vendor/laravel/framework/src/Illuminate/Database/Connection.php:829
  39  app/Utilities/Installer.php:161
```

The error message is misleading: the credentials are correct, and the database is reachable. The connection is simply not being made to the host and port that were given to the installer.

Running the installer a second time succeeds, because `.env` has been written in the meantime, so the correct values are read from the environment at boot. That makes the failure look intermittent.

### Cause

`Installer::saveDbVariables()` writes the values to `.env` and then updates the current connection, but the update does not reach the connection that is actually used:

- The `mysql` connection defines `read` and `write` host arrays (`config/database.php`), and those take precedence over the top level `host`. The existing `$db['host'] = $host;` line therefore has no effect on the connection.
- `port` is never assigned, so it keeps the value read from the environment at boot.

`isDbValid()` builds its own `install_test` connection with the given values, which is why validation passes immediately before the migration fails against a different host and port.

This affects any installation where the database is not on the host and port already present in `.env`: Docker, MAMP, cPanel, a secondary MySQL instance, or a remote server.

### Fix

Set `port` and the `read`/`write` hosts along with the values that were already being set.

### Steps to reproduce

1. Run a MySQL server on a non-default port, for example a Docker container published on `3307`
2. `cp .env.example .env`
3. `php artisan install --db-host=127.0.0.1 --db-port=3307 --db-name=akaunting --db-username=root --db-password=pass --admin-email=admin@company.com --admin-password=123456`

Before this change, the command exits with the error above and no tables are created. After it, the installation completes on the first run.

### Related

#3319 describes installing with a Docker MySQL container published on port `3307`, which is exactly the setup affected by this bug.

### DIFF
--- a/app/Utilities/Installer.php
+++ b/app/Utilities/Installer.php
@@ -223,6 +223,9 @@ class Installer
         $db = Config::get('database.connections.' . $con);
 
         $db['host'] = $host;
+        $db['port'] = $port;
+        $db['read']['host'] = [$host];
+        $db['write']['host'] = [$host];
         $db['database'] = $database;
         $db['username'] = $username;
         $db['password'] = $password;


cuneytsenturk:
Thank @Amadeus-22

PR #3391, opened 2026-08-30
Update docblocks left behind by the $job rename and route model binding

darkdi:
Most of these look like leftovers from renames. `Jobs` kept Laravel's `$command` after the argument became `$job`, while `dispatch` right above it already says `$job`. The two `Transfer` methods kept `$id` from before route model binding, and `edit` and `duplicate` in the same file are already updated. `landingPages` has a function call sitting where the type goes.


cuneytsenturk:
thanks for feedback @darkdi

PR #3394, opened 2026-09-01
Remove two array keys that a later duplicate already overrides

darkdi:
Two arrays set the same key twice. In `DocumentTransactions` the `payment` button has `'class' => 'long-texts'` and then `'class' => 'px-6 py-1.5 ... long-texts'` three lines later, and in `Version300` the permissions map lists `'settings-settings' => 'r,u'` on two consecutive lines. PHP keeps the last value, so the first `class` has never reached the button and the second permission line has never added anything.

Removed the losing line in each case, so nothing renders or seeds differently than it does today. The long class list already ends with `long-texts`, so the button keeps every class it has now.


cuneytsenturk:
Thank for all efforts @darkdi

PR #3395, opened 2026-09-07
fix: apply currency conversion when company currency matches account …

### LINKED ISSUE #3384: Currency conversion is not applied when company currency matches account currency
### Akaunting version

3.2.2

### PHP version

8.1.33

### Operating system

Windows 11

### Steps to reproduce

# Currency conversion is not applied when company currency matches account currency

## Description

When an invoice is issued in **GBP (£)** and a payment is made in **USD ($)**, the currency conversion works correctly when the company's currency differs from the account currency.

However, when the **company currency is the same as the account currency receiving the payment**, the conversion is not applied. The payment is recorded using the original invoice amount instead of the converted amount.

## Steps to reproduce

1. Create an invoice in **GBP (£)**.
2. Make a payment in **USD ($)**.
3. Set the currency rate to **1.33267**.
4. Compare the payment amount recorded in the account under different company/account currency configurations.

### Case 1 — Company currency: USD ($)

* Invoice currency: GBP (£)
* Payment currency: USD ($)
* Amount: **$253.00**
* Currency Rate: **1.33267**
* Amount credited to the account: **$253.00**
* **Currency conversion is not applied.**
* The Currency field does not work as expected.

### Case 2 — Company currency: GBP (£)

* Invoice currency: GBP (£)
* Payment currency: USD ($)
* Amount: **$337.17**
* Currency Rate: **1.33267**
* Amount credited to the account: **$337.17**
* **Currency conversion works correctly.**

### Case 3 — Company currency: EUR (€)

* Invoice currency: GBP (£)
* Payment currency: USD ($)
* Amount: **$337.17**
* Currency Rate: **1.33267**
* Amount credited to the account: **$337.17**
* **Currency conversion works correctly.**

## Expected result

The payment should be converted according to the specified currency rate regardless of whether the company currency matches the account currency.

For example, when the invoice/payment involves a currency conversion and the Currency Rate is **1.33267**, the converted amount should be calculated and recorded accordingly.

## Bug

The issue appears to occur specifically when:

**Company currency = Account currency**

In this situation, Akaunting skips the currency conversion and records the original payment amount instead.

### Additional comments

_No response_

### Relevant log output

```shell

```

coderlucifer:
Fixes #3384

## Problem

Currency conversion is skipped when the company's default currency matches the target currency in `convertBetween()`.

**Example from the issue:**
- Company currency: **USD**, Invoice: **GBP (£253)**, Payment: **USD**
- Expected: conversion applied → correct converted amount recorded
- Actual: conversion skipped → raw invoice amount recorded as-is

## Root Cause

`convertBetween()` converts amounts via a two-step pipeline through the company's default currency:

```
from_code → [Step 1: ÷ from_rate] → default_currency → [Step 2: × to_rate] → to_code
```

Step 1 has a guard — skip if `from_code == default_currency()` (amount is already in the default currency). But Step 2 unconditionally calls `convertFromDefault()` → `convert()`, which has its own short-circuit:

```php
if ($from == $to) {
    return $money->getAmount(); // Returns unchanged
}
```

When `to_code` equals the default currency, `convert()` sees `$from == $to` and returns the amount without applying the rate. The function is asymmetric: Step 1 is guarded but Step 2 is not.

## Fix

Added a symmetric guard for Step 2 in both **PHP** and **JavaScript** `convertBetween()`:

**PHP — `app/Traits/Currencies.php`:**
```diff
  if ($from_code != default_currency()) {
      $default_amount = $this->convertToDefault($amount, $from_code, $from_rate);
  }

- $converted_amount = $this->convertFromDefault($default_amount, $to_code, $to_rate);
+ if ($to_code != default_currency()) {
+     $converted_amount = $this->convertFromDefault($default_amount, $to_code, $to_rate);
+ } else {
+     $converted_amount = $default_amount;
+ }
```

**JavaScript — `resources/assets/js/views/common/documents.js`:**
```diff
  if (from_code != this.form.company_currency_code) {
      default_amount = this.convertToDefault(amount, from_code, from_rate, false, null);
  }

- let converted_amount = this.convertFromDefault(default_amount, to_code, to_rate, false, null);
+ let converted_amount;
+ if (to_code != this.form.company_currency_code) {
+     converted_amount = this.convertFromDefault(default_amount, to_code, to_rate, false, null);
+ } else {
+     converted_amount = default_amount;
+ }
```

When `to_code` IS the default currency, the amount after Step 1 is already correct — no further conversion needed. This makes the function logically symmetric and prevents the `convert()` short-circuit from silently dropping conversions.

## Impact

All call sites that use `convertBetween()` benefit from this fix automatically:
- `CreateBankingDocumentTransaction::checkAmount()`
- `UpdateBankingDocumentTransaction::checkAmount()`
- `Document::getPaidAttribute()`
- `Document::getReconciledAttribute()`
- `Transaction::getAmountForAccountAttribute()`
- `Transaction::getAmountForDocumentAttribute()`
- Frontend payment modal amount calculations

### DIFF
--- a/app/Traits/Currencies.php
+++ b/app/Traits/Currencies.php
@@ -48,7 +48,11 @@ trait Currencies
             $default_amount = $this->convertToDefault($amount, $from_code, $from_rate);
         }
 
-        $converted_amount = $this->convertFromDefault($default_amount, $to_code, $to_rate);
+        if ($to_code != default_currency()) {
+            $converted_amount = $this->convertFromDefault($default_amount, $to_code, $to_rate);
+        } else {
+            $converted_amount = $default_amount;
+        }
 
         return $converted_amount;
     }
--- a/resources/assets/js/views/common/documents.js
+++ b/resources/assets/js/views/common/documents.js
@@ -281,7 +281,12 @@ const app = new Vue({
                 default_amount = this.convertToDefault(amount, from_code, from_rate, false, null);
             }
 
-            let converted_amount = this.convertFromDefault(default_amount, to_code, to_rate, false, null);
+            let converted_amount;
+            if (to_code != this.form.company_currency_code) {
+                converted_amount = this.convertFromDefault(default_amount, to_code, to_rate, false, null);
+            } else {
+                converted_amount = default_amount;
+            }
 
             return converted_amount;
         },


cuneytsenturk:
Hi @coderlucifer, and thank you for taking the time on this one — the issue you picked up (#3384) is real and worth fixing.

I'm going to close this PR without merging, though, because the change ends up being a no-op. `convertFromDefault()` delegates to `convert()`, and `convert()` already short-circuits when the source and target currencies are the same:

```php
// app/Traits/Currencies.php:13
// No need to convert same currency
if ($from == $to) {
    return $format ? $money->format() : $money->getAmount();
}
```

Since `convertFromDefault()` passes the default currency as `$from` (`app/Traits/Currencies.php:40`), the case your `else` branch handles — `$to_code` being the default currency — is already handled one level down. The JS side has the identical guard at `resources/assets/js/views/common/documents.js:248`, so the same applies there.

There is one behavioural difference, but it goes the wrong way: on the existing path the amount passes through `Money`, which rounds it to the currency's precision, while the new `else` branch returns the raw value. So the patch would quietly change rounding behaviour without fixing anything.

More importantly, it does not cover the scenario from the issue. In Case 1 (company USD, invoice GBP, payment USD) `$to_code` is GBP and the default currency is USD, so the branch you added is never entered — the result is identical before and after.

So the bug is real but the root cause is elsewhere. I'm actively working on it, most likely around how the transaction's `currency_code` and `currency_rate` are recorded and read back rather than in `convertBetween()` itself.

Please don't let this discourage you — your other PR (#3396) was spot on and is already merged. Thanks again for contributing.

coderlucifer:
Thanks for the detailed explanation, @cuneytsenturk !

You make a great point about the Money formatting and precision rounding.. I completely missed that the else branch would bypass it and return a raw float.

While digging deeper into this, I also noticed that the payment modal `payment.blade.php` initializes the hidden `currency_code` field to the document's currency, and the JS `onChangePaymentAccount` only updates it if the user actually changes the dropdown. So I completely agree that the root cause is upstream in how those values are recorded initially.

Thanks for taking the time to review it so thoroughly and explain the architecture. I learned a lot from this one! Glad the search filter PR was helpful, and I'm looking forward to picking up another issue soon.

